PR #2, opened 2017-01-25
Implemented DES

gsingh93:
This is an implementation of DES. The included benchmark gives 2 MB/s performance, I'm not sure if that's good or bad. Most of the trickiness with the implementation is the fact that the most significant bit is considered bit zero instead of the least significant bit. It's possible I did some unnecessary shifting/arithmetic to account for this.

Note that the crate name `des` is taken on crates.io, so we might have to change that name.

I'm planning on adding more test cases. I have a file of `(input, key, output)` tuples that I need to parse and turn into the correct format for tests used in this repo.


gsingh93:
Hmm, now that I think about it, it would probably be cleaner to just use big endian integers and work from the LSB instead of using little endian and working from the MSB. I'll experiment with that.

newpavlov:
Thank you!
I will review your PR on this weekend. Regarding performance, this [paper](http://www.cse.wustl.edu/~jain/cse567-06/ftp/encryption_perf/) gives DES 21 MB/s, so there is room for improvement. You can use [`perf` and flamegraphs](http://blog.adamperry.me/rust/2016/07/24/profiling-rust-perf-flamegraph/) for finding hot spots, additionally try to play with `#[inline(always)]` and loop unrolling, they can significantly improve performance.

Also you can implement 3DES, not sure if it's better to use the separate crate for it or include in this one.

newpavlov:
Also check this [implementation](https://github.com/antoyo/des-rs), I will try to contact owner of this crate later to ask if he would like to cooperate with us.

newpavlov:
I've updated your code with optimizations I borrowed from antoyo's `des-rs`, it makes code less readable, but significantly faster. It could be nice to add tests which will check optimized transformations against straightforward ones written by you. I did similar [test](https://github.com/RustCrypto/block-ciphers/blob/master/magma/src/gen_table.rs) in `magma`. Also I've moved subkeys generation into `new()`, so if state will be reused (for whatever reason) subkeys will not be generated again. I'll add benchmarks with cipher initialization on each block later, as currently it benchmarks only encryption without state initialization.

Currently I am getting 17 MB/s without initialization and 11 MB/s with it. Meanwhile OpenSSL implementation gives approximately 50 MB/s, so there is still room for improvement.

des/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+documentation = "https://docs.rs/des"
+repository = "https://github.com/RustCrypto/block-ciphers"
+keywords = ["crypto", "des", "block-cipher"]
+
+[dev-dependencies]
+crypto-tests = "0.2"
+
+[dependencies]
+generic-array = "0.5"
+block-cipher-trait = "0.2"
```
newpavlov: You use old versions of crates:
```
generic-array = "0.6"
block-cipher-trait = "0.3"
```
(Forgot to update blowfish)

des/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    fn decrypt_block(&self, input: &Block<U8>, output: &mut Block<U8>) {
+        let block = read_u64_be(input);
+        let res = self.decrypt(block);
+        write_u64_be(output, res);
+    }
+}
+
+impl BlockCipherFixKey for Des {
+    type KeySize = U8;
```
newpavlov: Shouldn't it be `U7`?

des/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+impl BlockCipher for Des {
+    type BlockSize = U8;
+
+    fn encrypt_block(&self, input: &Block<U8>, output: &mut Block<U8>) {
+        let block = read_u64_be(input);
+        let res = self.encrypt(block);
+        write_u64_be(output, res);
```
newpavlov: I've added `read_u64_be` to `byte-tools`. Don't forget to run `cargo update` to get new version.

antoyo:
I've been invited to participate to this discussion, so here I am :p .
I won't have much time to work on this, unfortunately, but I can give you access to the `des` crate on crates.io.
I don't mind adding the Apache license.

How can I give you a write access to the `des` crate?
Is it throught the crates.io website?
I cannot access it anymore: the login doesn't work for me in Firefox and the other browsers that I use.
I already had this issue in the past, but I don't remember what was the fix.
Is it related to third-party cookies (I tried disabling them)?

newpavlov:
@antoyo 
Thank you for your participation!

>How can I give you a write access to the des crate?

You can do it using cargo with this command: `cargo owner --add newpavlov`. More information about it available [here](http://doc.crates.io/crates-io.html#cargo-owner).

>I cannot access it anymore: the login doesn't work for me in Firefox and the other browsers that I use.

You don't need to be authenticated on crates.io if your cargo token is intact. Your problems with cargo.io could be because of privacy settings, as it uses github for authentication. As a backup plan you can contact [Alex Crichton](mailto:acrichton@mozilla.com).

antoyo:
I added you as an owner.
Thanks for your interest in my work.
I can answer any question about the implementation.

newpavlov:
@antoyo 
Thank you! I will publish new version after merge.

Meanwhile can you compare your S-box implementation with gsingh93's [implementation](https://github.com/RustCrypto/block-ciphers/pull/2/files#diff-a9462739b84c7b8f90cc5ec8f05c35e3R154). As I understand in the later transformed S-boxes are used to improve performance. Simple tests show that this implementation is a bit faster than yours, but maybe you know about additional optimizations which can be applied here?

newpavlov:
Sorry it took so long!

I will merge this PR, although before publishing I'll need to make some changes.

newpavlov:
BTW it looks like you had a bug in decryption part.

newpavlov:
@gsingh93 
If you'll have time please take a look at #12.

gsingh93:
Ack. Fairly busy for a while, so if you or anyone can fix that would be great. Otherwise, I'll take a look when I have time.

PR #4, opened 2017-04-23
RC2 cipher

Trojan295:
I've created the RC2 cipher. The benchmark shows a throughput of 68 MB/s. Two test vectors are included.


newpavlov:
Thank you for your contribution!

Your code looks good, but unfortunately it does not pass tests 4, 5, 6 and 8 from [RFC 2268](https://www.ietf.org/rfc/rfc2268.txt). Seems problem lies somewhere in the key expansion. You can download test files [here](http://rgho.st/download/7SbJyLtWY/c10b0b6c3fc528ac8a14d1ec5610c3f55bd9ce62/data.tar.gz).

Trojan295:
I see that all test that fail have a different "effective key length" than the real key length. Currently I assumed, that the effective key length is simply the length of the key. Test 6 and 7 are the same expect the effective key length, so looks this should be a parameter to the cipher.
I'm not sure how to implement this. The BlockCipherVarKey trait has only the key as a parameter.
Besides, how does providing a different effective length affect the performance and security of the cipher? Perhaps we could for now don't parametrize this.

newpavlov:
You can add something like `pub fn new_with_effective_key(key: &[u8], effective_key: usize) -> Self` into RC2 `impl`. This will not be covered by the trait, but it's all right. Also `BlockCipherVarKey` can be defined in terms of this function with `effective_key = key.len()`. One drawback though, you'll need to write/copy test function, as `crypto-tests` does not cover such cases.

Trojan295:
I added all those test vectors from RFC 2268 and now they pass. I've done this how you suggested, by adding `pub fn new_with_effective_key_length(key: &[u8], eff_key_len: usize) -> RC2`. The constructor in `BlockCipherVarKey` assumes, that the effective length is the length of the provided key.

rc2/tests/test_vectors.rs
```diff
@@ -0,0 +1,38 @@
+
+pub struct RC2CipherTest {
+    pub key: &'static [u8],
+    pub eff_key_length: usize,
```
newpavlov: I think it's better to store tests as files and use `new_block_cipher_tests`, so code will look something like this:
```Rust
#[test]
fn rc2_effective_key_64() {
    let mut buf = GenericArray::new();
    let tests = new_block_cipher_tests!("3", "4", "5", "6");
    for test in tests {
        let cipher = rc2::RC2::new_with_effective_key_length(test.key, 64);

        cipher.encrypt_block(&from_slice(test.input), &mut buf);
        assert_eq!(test.output, &buf[..]);
        cipher.decrypt_block(&from_slice(test.output), &mut buf);
        assert_eq!(test.input, &buf[..]);
    }
}
```

Same for 129 bits. This way all tests will be stored in the same format. If you don't want to deal with the boilerplate, you can write a simple macro. Also you can add sub-folders, say, `data/eff_key_64/` and `data/eff_key_129/` if you want to distinguish different types of tests.

Also you missed test with the single byte key. (number 4)

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn new_with_effective_key_length(key: &[u8], eff_key_len: usize) -> RC2 {
+        RC2 {
+            exp_key: RC2::expand_key(key, eff_key_len)
+        }
+    }
+
+    fn expand_key(key: &[u8], t1: usize) -> [u16; 64] {
+        let key_len = key.len() as usize;
+
+        //let t1: usize = key_len<<3;
```
newpavlov: Please remove this line.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        ct[0] = r[0] as u8;
+        ct[1] = (r[0] >> 8) as u8;
+        ct[2] = r[1] as u8;
+        ct[3] = (r[1] >> 8) as u8;
+        ct[4] = r[2] as u8;
+        ct[5] = (r[2] >> 8) as u8;
+        ct[6] = r[3] as u8;
+        ct[7] = (r[3] >> 8) as u8;
+
+        ct
```
newpavlov: You can construct and return this array immediately like this:
```Rust
[
    r[0] as u8,
    (r[0] >> 8) as u8,
    ...
]
```

Same can be done at the lines 107-112 for construction of `r`. Initialization is probably optimized away, so performance wise they are the same, but it will make code a bit shorter.

Later I will probably add `u16` methods to the `byte-tools` crate, so it could be written even shorter.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        r[0] = r[0].wrapping_sub(self.exp_key[(r[3] & 63) as usize]);
+    }
+
+    fn encrypt(&self, block: [u8; 8]) -> [u8; 8] {
+        let mut r: [u16; 4] = [0; 4];
+
+        r[0] = ((block[1] as u16) << 8) + (block[0] as u16);
+        r[1] = ((block[3] as u16) << 8) + (block[2] as u16);
+        r[2] = ((block[5] as u16) << 8) + (block[4] as u16);
+        r[3] = ((block[7] as u16) << 8) + (block[6] as u16);
```
newpavlov: See comment regarding `ct`.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        //let t1: usize = key_len<<3;
+        let t8: usize = (t1+7)>>3;
+
+        let tm: usize = (255 % ((2 as u32).pow((8+t1-8*t8) as u32))) as usize;
+
+        let mut key_buffer: [u8; 128] = [0; 128];
+        for i in 0..key_len {
+            key_buffer[i] = key[i];
+        }
```
newpavlov: Maybe use `copy_from_slice` instead of explicit loop?

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        ct
+    }
+
+    fn decrypt(&self, block: [u8; 8]) -> [u8; 8] {
+        let mut r: [u16; 4] = [0; 4];
+
+        r[0] = ((block[1] as u16) << 8) + (block[0] as u16);
+        r[1] = ((block[3] as u16) << 8) + (block[2] as u16);
+        r[2] = ((block[5] as u16) << 8) + (block[4] as u16);
+        r[3] = ((block[7] as u16) << 8) + (block[6] as u16);
```
newpavlov: See comment about `encrypt`.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        ct[0] = r[0] as u8;
+        ct[1] = (r[0] >> 8) as u8;
+        ct[2] = r[1] as u8;
+        ct[3] = (r[1] >> 8) as u8;
+        ct[4] = r[2] as u8;
+        ct[5] = (r[2] >> 8) as u8;
+        ct[6] = r[3] as u8;
+        ct[7] = (r[3] >> 8) as u8;
+
+        ct
```
newpavlov: See comment about encrypt.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl BlockCipherVarKey for RC2 {
+    fn new(key: &[u8]) -> RC2 {
+        RC2 {
+            exp_key: RC2::expand_key(key, key.len()*8)
+        }
```
newpavlov: You can also write it as just `RC2::new_with_effective_key_length(key, key.len()*8)`. But it's matter of preference.

rc2/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl BlockCipher for RC2 {
+    type BlockSize = U8;
+
+    fn encrypt_block(&self, input: &Block<U8>, output: &mut Block<U8>) {
+        let mut iblock: [u8; 8] = [0; 8];
+        for i in 0..8 {
```
newpavlov: This method can be shortened to just one line `self.encrypt(input)` if you'll add changes described in the comment to `encrypt` method. Same goes for `decrypt_block` and `decrypt`.

rc2/src/lib.rs
Trojan295: After doing these changes `cargo bench` on my PC shows:

> test encrypt ... bench:         122 ns/iter (+/- 3) = 65 MB/s

and my previous solution (create array with zeros and fill them afterwards)

> test encrypt ... bench:         116 ns/iter (+/- 9) = 68 MB/s

Looks the compiler behaves a bit different in these cases.

rc2/src/lib.rs
Trojan295: Hmm.. The `encrypt` function in `RC2` is taking an array and returning an array, while in `trait BlockCipher` it's outputing to a provided pointer, so it cannot be plug in as one line without changing the definiton of `RC2::encrypt`. I tried to go further and implement the whole algorythm in `encrypt_block` like:

    fn encrypt_block(&self, input: &Block<U8>, output: &mut Block<U8>) {
        let mut r: [u16; 4] = [
        ((input[1] as u16) << 8) + (input[0] as u16),
        ((input[3] as u16) << 8) + (input[2] as u16),
        ...
        output[6] = r[3] as u8;
        output[7] = (r[3] >> 8) as u8;
    }

This this causes an even bigger drop in the performance on my PC:

> test encrypt ... bench:         143 ns/iter (+/- 3) = 55 MB/s

It it possible, that `Block<U8>` adds some additional abstration, which causes that operations on it are slower than on generic Rust arrays?

newpavlov: This is definitely strange, as I understand `generic-array`, `Block<U8>` should compile into the same code as for `[u8; 8]`. I will try to experiment with this. Hope I'll find the problem.

newpavlov:
I've pushed minor modifications. On my machine they improve performance approximately by 10%. Please check it on your machine.

If you don't have anything to change we are good to merge!

Trojan295:
Looks fine from my side. The performance is comparable to my first solution.

Can be merged.

newpavlov:
Thank you for your work again!

PR #5, opened 2017-05-05
Added threefish

newpavlov:
Code based on cbarcenas's [implementation](https://github.com/DaGenix/rust-crypto/pull/400) with his consent.


Slals:
Out of curiosity, it seems you forced-push treefish but nowhere to be found in master @newpavlov, any reason?

Nakilon:
How do I used it? Never used Rust.
I've installed `cargo`, `cargo show` and it pointed at this repo. Now what? I would like to use it in a terminal to encrypt files with threefish-256.

PR #7, opened 2017-06-11
Twofish cipher

labdsf:
Do not merge it now, it is broken.

I created pull request just to notify that I have something working so you don't duplicate effort.

I took test from
https://www.schneier.com/academic/paperfiles/paper-twofish-paper.pdf

It should output `9F589F5CF6122C32B6BFEC2F2AE8C35A`, I got the same bytes but endianess looks wrong. Also no decrypt yet and BlockCipher is incomplete.

The code is loosely based on libtomcrypt which is public domain / WTFPL


newpavlov:
Hi! Welcome to the project!

You can take scottm73's [implementation](https://github.com/DaGenix/rust-crypto/pull/361) as a reference if you want.

Also I plan work on block modes soon, what do you need for passwordsafe?

labdsf:
> Also I plan work on block modes soon, what do you need for passwordsafe?

CBC and ECB only

labdsf:
@newpavlov PR is ready for merge

twofish/LICENSE-APACHE
```diff
@@ -0,0 +1,13 @@
+Copyright 2017 Damian Czaja
```
newpavlov: I think it's better to take LISENCE-APACHE which is use across other crates. (e.g. from streebog) I've missed that rc2 crate uses a different file.

twofish/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2017 Damian Czaja
```
newpavlov: It's better to place your name here. :)

twofish/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+             0x7C6CF0C4, 0x2F9109C8, 0xE69EA8D1, 0xED99BDFF, 0x35DC0BBD, 0xA03E5018, 0xFB18EA0B,
+             0x38BD43D3, 0x76191781, 0x37A9A0D3, 0x72427BEA, 0x911CC0B8, 0xF1689449, 0x71009CA9,
+             0xB6363E89, 0x494D9855, 0x590BBC63, 0xF95A28B5, 0xFB72B4E1, 0x2A43505C, 0xBFD34176,
+             0x5C133D12, 0x3A9247F7, 0x9A3331DD, 0xEE7515E6, 0xF0D54DCD];
+        assert_eq!(&twofish.s[..], &sboxkey[..]);
+        assert_eq!(&twofish.k[..], &subkeys[..]);
+    }
+
+    #[test]
+    fn encrypt_ecb128() {
```
newpavlov: Use please macros from `crypto-tests` for it and place it in the `tests/` directory. (take other ciphers as an example) Same goes for `encrypt_ecb192` and `encrypt_ecb256`.

twofish/LICENSE-APACHE
labdsf: fixed

twofish/LICENSE-MIT
labdsf: fixed

twofish/src/lib.rs
labdsf: Moved integration tests to tests/, but not sure if crypto-tests library is useful here.
I copied tests from Twofish paper. All tests are related: plaintext is used as a key for next test, and ciphertext is used as a plaintext. And it is easier to compare to original paper when tests are not binary.

newpavlov: You are right, one small thing you could change though is to use `match` instead of chained `if else`:

```Rust
let correct = match i {
    1 => Some([...]),
    2 => Some([...]),
    ...
    _ => None,
};
correct.map(|v| assert_eq!(cipher.as_slice(), v));
```

I think it will improve your code.

Also please do not forget about benchmarks.

labdsf:
Added benchmarks

newpavlov:
I would like to change some minor things, but overall it looks good, so I will merge it. (but still no benchmarks, I will add them later ;) ) First version of blockmodes should be ready next week.

Though I think your implementation can be substantially optimized by using precomputed tables, as currently it gives 2 MB/s on my machine. See "[Optimized C Implementation](https://www.schneier.com/code/twofish-optimized-c.zip)" for the reference, if you'll have time to write more optimized version.

labdsf:
> but still no benchmarks, I will add them later ;)

I forgot to do `git add`, submitted a separate PR #8

PR #8, opened 2017-06-18
Add Twofish benchmarks

labdsf:
Forgot to do `git add`


newpavlov:
Sorry it got delayed this way, I've pushed update which depends on new version of `crypto-tests` and contains benchmarks. So I'll close this PR.

PR #11, opened 2017-11-05
Update generic-array to 0.9

roblabla:
Necessary since https://github.com/RustCrypto/traits/commit/4e4ecaf0c951995525570ed1212d43fe56c2ae7a


newpavlov:
I've planned to do it after finishing with hashes, but you did it before me. :)

Could you please use [re-exported](https://github.com/RustCrypto/traits/blob/master/block-cipher-trait/src/lib.rs#L4) version of `generic-array` from `block-cipher-trait` crate? This way we will be able to remove explicit import from `Cargo.toml`.

roblabla:
Is this fine ? I used a sed to refactor everything.

newpavlov:
Yes, looks good! Thank you!

PR #13, opened 2017-11-18
Add new_from_cipher() constructor to CtrAes*

tarcieri:
This is the API I would prefer for instantiating `CtrAes*`. I already have my own wrappers for the `Aes*Ni` types that have done key expansion up front which I'd like to reuse to instantiate `CtrAes*` with.

Ideally I think this would be the type signature for `new` but I figured I'd go for a simple additive change first (especially as I'd like to start working off this branch in Miscreant)

### DIFF
--- a/aesni/src/ctr.rs
+++ b/aesni/src/ctr.rs
@@ -41,6 +41,15 @@ macro_rules! impl_ctr {
                 }
             }
 
+            pub fn new_from_cipher(cipher: $cipher, nonce: &[u8; BLOCK_SIZE]) -> Self {
+                let ctr = u64x2::read(nonce).swap_bytes();
+                Self{
+                    ctr, cipher,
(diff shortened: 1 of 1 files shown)


newpavlov:
Is there a particular reason why you prefer to initialize block mode from initialized cipher and not from the key? If we add reset functionality to block mode and stream ciphers shouldn't it cover all reasonable use-cases? For example if we consider cipher-based MACs you'll use two different keys for MAC function and for encryption, so you can't reuse initialized cipher.

tarcieri:
Yes, reset functionality would accomplish the same thing

newpavlov:
Relevant issue for further discussion: https://github.com/RustCrypto/traits/issues/6

tarcieri:
Subscribed!

PR #14, opened 2017-12-02
Threefish: fixed for updated block-cipher-traits

kazcw:


kazcw:
I no longer have time to keep synchronizing this PR with the published crate. Threefish can be found here: https://github.com/cryptocorrosion/cryptocorrosion/tree/master/block-ciphers/threefish

PR #17, opened 2018-03-19
Ctr trait

tarcieri:
Provides an `xor()` method similar to the one(s) that used to exist in the aesni crate. Does not pad and handles messages which aren't aligned to the cipher's underlying block size.

Could definitely use some tests, but I thought I'd throw this up first as a strawman.


newpavlov:
Solved by `ctr` and `aes-ctr` crates in the RustCrypto/stream-ciphers.

PR #22, opened 2018-08-07
Add feature to disable aesni target feature compile time checks

cheme:
When using runtime `is_x86_feature_detected` macro to check if I could use aesni, I had to disable the checks.
Is it possible to include this little feature to allow usage of aesni without compile time checks?

### DIFF
--- a/aes/aesni/Cargo.toml
+++ b/aes/aesni/Cargo.toml
@@ -21,6 +21,7 @@ stream-cipher = { version = "0.1", features = ["dev"] }
 [features]
 default = ["ctr"]
 ctr = []
+nocheck = []
 
 [workspace]
 
--- a/aes/aesni/src/lib.rs
+++ b/aes/aesni/src/lib.rs
@@ -68,7 +68,7 @@ pub extern crate block_cipher_trait;
 #[macro_use] extern crate opaque_debug;
 #[cfg(feature = "ctr")]
 pub extern crate stream_cipher;
-
+#[cfg(not(feature = "nocheck"))]
 mod target_checks;
 #[macro_use]
 mod utils;


newpavlov:
Problem with such feature is that intrinsics will not get inlined in such setup, which will severely impact performance. I am not sure if `#[target_feature(enable = "aes")]` on your function will help here, as I think methods from `aesni` crate will be compiled first.

I guess this PR can be merge for now, but ideally we need the language support for such use-cases.

@gnzlbg
Do you have ideas what we can do here?

gnzlbg:
The problem is that this crate requires `aes` to be enabled at compile-time. That obviously makes the crate unusable for people that want to use it if the CPU supports `aes` at run-time, but IIUC this is by design. 

I don't really understand why it was designed this way.

A better design would be to have the crate just use `#[target_feature(enable = "aes")]` and provide an unsafe API. 

This allows those who want to do run-time feature detection for `aes` to use this crate by just writing: `if is_x86_feature_detected!("aes") { unsafe { aesni::foo(...) } }` - note that this won't do run-time feature detection _iff_ `aes` is enabled at compile-time.

This also allows those using compile-time feature detection to use the crate by just writing: `#[cfg(target_feature = "aes")] { unsafe { aesni::foo(...) } }`.

You can then provide safe APIs on top of this in this same crate or other crates. For example:

```rust
#[cfg(target_feature = "aes")] 
pub mod ct { // compile-time 
    pub fn foo() { unsafe { ::foo() } }
}

pub mod rt { // run-time
   pub fn foo() {
        if is_x86_feature_detected!("aes") { 
            unsafe { ::foo() }
        } else {
            // panic!();
            // fallback_foo();
            // Err(...) ...
        }
   }
}
```

> Problem with such feature is that intrinsics will not get inlined in such setup, 

If a user writes in their own crate:

```rust
fn my_alorithm() {
    if is_x86_feature_detected!("aes") {
        unsafe { foo() }
    } 
}

#[target_feature(enabled = "aes")] 
fn foo() {
    // calls to aesni functions
}
```

All `aesni` functions that have `#[target_feature(enabled = "aes")]` and `#[inline]` will be inlined into `foo` unless inlining is not profitable.

newpavlov:
>I don't really understand why it was designed this way.

The main reason for the current design is that block ciphers expose their functionality via `BlockCipherTrait` methods of which are safe. Another reason is that I think that current `unsafe` approach to the target features is a wrong solution to the problem. (albeit the easiest one to implement and stabilize) This is why I've wrote a Pre-RFC for [target restriction contexts](https://internals.rust-lang.org/t/pre-pre-rfc-target-restriction-contexts/7163), but it looks like improving target features handling is not a priority for Rust team right now. Also I don't think that `is_x86_feature_detected!` should be used on block cipher level, ideally this check should be pushed as higher level as possible, but it contradicts to the modular design of RustCrypto. (and language does not provide any tools for having several crate versions with different features enabled)

I guess as a stop-gap solution I could expose trait-less feature-gated unsafe methods with `#[target_feature(enabled = "aes")]`.

gnzlbg:
> I guess as a stop-gap solution I could expose trait-less feature-gated unsafe methods with #[target_feature(enabled = "aes")].

That is probably a good step forward. You can reuse these in the safe ones by just calling them via an `unsafe { }`. 

> Also I don't think that is_x86_feature_detected! should be used on block cipher level, ideally this check should be pushed as higher level as possible, but it contradicts to the modular design of RustCrypto. 

There is a "pattern" in the wild which allows you to perform run-time feature detection yourself.

```rust
[derive(Clone, Copy, Debug)]
pub struct AESNIBuilder(());

impl AESNI {
    pub fn new() -> Option<AESNIBuilder> {
        if is_x86_feature_detected!("aes") {
            Some(AESNIBuilder(()))
        } else {
            None
        }
    }
    #[inline] // maybe #[target_feature(enable = "aes")] + unsafe fn
    fn foo() { unsafe { aesni::foo() } }
}
```

People are using it with `#[inline]` and without `#[target_feature]` with success, but even if you have to make it `#[target_feature]`+`unsafe` for it to inline properly, you know that calling all these methods is safe if you have an `AESNIBuilder`. 

Then, whether the user wants to construct a new builder every time, or put one in the whole program behind a static, that kind of is up to them.

newpavlov:
>Arguably, having to make your APIs unsafe while using #[target_feature] is annoying, but that is currently how things are. The target_feature 1.1 RFC fixes this (rust-lang/rfcs#2396), but it hasn't been merged yet.

It's not only about annoyance of `unsafe`. I think ideally language should support compiling several versions of a function with different sets of enabled target features. (I've heard gcc supports something similar?) This way crates will be able to use a simple compile-time feature detection without builders bloat, while users will be able to use run-time detection with several hierarchies if they want to. But drawback can be a binary bloat and increased compilation times (as you'll have to compile several versions of upstream dependencies), especially if done naively.

gnzlbg:
> I think ideally language should support compiling several versions of a function with different sets of enabled target features.

The language already supports this.

gnzlbg:
> My wording was a bit unclear, I've meant that if you'll use function foo from crate bar with enabled target features, then the whole crate will be compiled with those features, so compile time checks will not create problems for users who'll use run-time detection and #[target_feature(enable = "..")].

I still don't understand what you mean here. 

> if you'll use function foo from crate bar with enabled target features,

Enabled how?

newpavlov:
>I still don't understand what you mean here.

With `#[target_feature(enable = "..")]`.

The system which I've tried to describe will be able to compile the following code:
```Rust
#[target_feature(enable = "aes")]
unsafe fn aesni_encrypt(data: &mut [u8]) {
    // use aesni crate here
}

if is_x86_feature_detected!("aes") { 
    unsafe { aesni_encrypt(block) };
} else {
    // software AES
    soft_encrypt(block);
}
```
Because `aesni` used only in the context with enabled `aes` target feature, `aesni` compile-time checks will pass, as compiler will build only one version of `aesni` with the enabled target feature.

If somewhere in the code `aesni` crate will be used in the context with two enabled features, say `aes` and `sse3`, then two versions of `aesni` crate will be compiled for both sets of target features.

gnzlbg:
> The system which I've tried to describe will be able to compile the following code:

But that code already works in Rust, just as you wrote it. 

> Because aesni used only in the context with enabled aes target feature, aesni compile-time checks will pass, as compiler will build only one version of aesni with the enabled target feature.
>
> If somewhere in the code aesni crate will be used in the context with two enabled features, say aes and sse3, then two versions of aesni crate will be compiled for both sets of target features.

Wouldn't that break separate compilation, in that you would have to wait till you use a crate to know how to compile it? That is, you can't compile your main crate because your dependencies haven't been compiled, yet you cannot compile your dependencies because you don't know how they are going to be used and therefore you don't know which target features you have to enable to compile them?

newpavlov:
>But that code already works in Rust, just as you wrote it.

No, with the current version of `aesni` it will not work, because it checks if target features are enabled in the very [beginning](https://github.com/RustCrypto/block-ciphers/blob/master/aes/aesni/src/lib.rs#L72) of the crate. This check allows me to be sure that necessary target features are enabled. Thus when you try to compile it, even if you use crate in contexts with enabled `aes` it will still produce compile time error. Manually disabling this check via feature as was done in this PR will allow to compile the snippet, but it will mark "unsafe" code (in the sense of the used intrinsics) as a completely safe one.

>Wouldn't that break separate compilation, in that you would have to wait till you use a crate to know how to compile it?

I guess, yes, it will. Technically it's possible to work around this issue (e.g. by compiling crate without features first and postponing any errors until crate will be used in feature-less context), but overall I agree it was not the best idea.

gnzlbg:
> No, with the current version of aesni it will not work, because it checks if target features are enabled in the very beginning of the crate. 

I think I missed the `// use aesni crate here` part.

I think I missed that. It makes sense now.

> This check allows me to be sure that necessary target features are enabled.

While you could use `cfg(target_feature)` for that, it will also check that the **whole binary** is compiled with a target feature enabled. If you just want to force a single function to be compiled with some target features enabled, that's what `#[target_feature]` is for. 

> Technically it's possible to work around this issue (e.g. by compiling crate without features first and postponing any errors until crate will be used in feature-less context)

Features affect everything since you can do conditional compilation with `cfg`, so they can affect path resolution, macro expansion, type checking, ... For example, in the case of `aesni`, the crate doesn't compile without features, so trying to compile it "first" somehow, won't give you any information, and even if it would give you any information, that could completely change once the target features change because the whole crate can change.

So at least with the current system, I don't think this is technically possible. One would need probably many language features and technical work to come up with a system here that makes sense and works as the user expects.

newpavlov:
>If you just want to force a single function to be compiled with some target features enabled, that's what #[target_feature] is for.

But it's not a single functions, all functions and structs in the crate require enabled `aes` target feature (and `ssse3` if you'll enable `ctr` feature) and want to provide safe interface (so target_feature v1.1 will not help here much), so it's essentially something like wanting working `#![target_feature]`.

I guess for now we will have to use feature for disabling target feature checks. Personally I don't think that builder pattern proposed earlier is a good solution to the problem in this particular case.

@cheme 
I'll merge this PR and after doc improvement with CI fix I'll publish an update to crates.io.

PR #24, opened 2018-08-07
More pedantic unsafe code

### LINKED ISSUE #23: Too much undefined behavior
I started to write a PR to just show how using `#[target_feature]` would look like but I encountered too much undefined behavior in the process (fixing it was taking more time than actually dealing with target feature). 

I've only taken a look at a tiny fraction of the `aesni` crate, but it seems that references to uninitialized memory are created pretty much everywhere (all references must point to initialized memory [0]): 
    * https://github.com/RustCrypto/block-ciphers/blob/master/aes/aesni/src/aes128/expand.rs#L21
    * https://github.com/RustCrypto/block-ciphers/blob/master/aes/aesni/src/aes128/expand.rs#L36

The way I fixed these was by using `ptr::write` to write to uninitialized memory without creating a reference. 

There are also reads memory out-of-bounds: https://github.com/RustCrypto/block-ciphers/blob/master/aes/aesni/src/aes192/expand.rs#L51

The way I fixed this was by creating a stack buffer of appropriate size in the `mod.rs` file which I initialized with `mem::uninitialized`, and then using `ptr::write` to write the `GenericArray` into the head of the buffer, so that the vector load never reads memory out of bounds - the memory read is only partially initialized though, and reading vector lanes with uninitialized memory should be done with care (so that no UB is introduced). 

---

[0] `&T as *const T` and `&mut T as *mut T` might not end up being undefined behavior because they _might_ end up being atomic operations in the memory model that create a raw pointer directly without creating a reference (so that no reference to uninitialized memory is ever created by these expressions, preventing them from invoking undefined behavior).

newpavlov:
Fixes #23 

@gnzlbg
What do you think? Do you have any suggestions where unsafe code can be improved? I will not be using `union` approach here yet, maybe I'll change my mind in future.

### DIFF
--- a/aes/aesni/src/aes192/expand.rs
+++ b/aes/aesni/src/aes192/expand.rs
@@ -54,7 +54,7 @@ pub(super) fn expand(key: &[u8; 24]) -> ([__m128i; 13], [__m128i; 13]) {
         // note that unaligned load MUST be used here, even though we read
         // from the array (compiler missoptimizes aligned load)
         let (k0, k1l) = {
-            let mut t: [u8; 32] = mem::uninitialized();
+            let mut t = [0u8; 32];
             ptr::write(t.as_mut_ptr() as *mut [u8; 24], *key);
             (
                 _mm_loadu_si128(t.as_ptr() as *const __m128i),
--- a/aes/aesni/src/aes192/expand.rs
+++ b/aes/aesni/src/aes192/expand.rs
@@ -1,91 +1,107 @@
 use arch::*;
 
-use core::mem;
+use core::{mem, ptr};
 
 macro_rules! expand_round {
-    ($t1:expr, $t3:expr, $round:expr) => {
-        unsafe {
-            let mut t1 = $t1;
-            let mut t2;
-            let mut t3 = $t3;
-            let mut t4;
+    ($t1:expr, $t3:expr, $round:expr) => {{
+        let mut t1 = $t1;
+        let mut t2;
+        let mut t3 = $t3;
+        let mut t4;
 
-            t2 = _mm_aeskeygenassist_si128(t3, $round);
-            t2 = _mm_shuffle_epi32(t2, 0x55);
-            t4 = _mm_slli_si128(t1, 0x4);
-            t1 = _mm_xor_si128(t1, t4);
-            t4 = _mm_slli_si128(t4, 0x4);
-            t1 = _mm_xor_si128(t1, t4);
-            t4 = _mm_slli_si128(t4, 0x4);
-            t1 = _mm_xor_si128(t1, t4);
-            t1 = _mm_xor_si128(t1, t2);
-            t2 = _mm_shuffle_epi32(t1, 0xff);
-            t4 = _mm_slli_si128(t3, 0x4);
-            t3 = _mm_xor_si128(t3, t4);
-            t3 = _mm_xor_si128(t3, t2);
+        t2 = _mm_aeskeygenassist_si128(t3, $round);
+        t2 = _mm_shuffle_epi32(t2, 0x55);
+        t4 = _mm_slli_si128(t1, 0x4);
+        t1 = _mm_xor_si128(t1, t4);
+        t4 = _mm_slli_si128(t4, 0x4);
+        t1 = _mm_xor_si128(t1, t4);
+        t4 = _mm_slli_si128(t4, 0x4);
+        t1 = _mm_xor_si128(t1, t4);
+        t1 = _mm_xor_si128(t1, t2);
+        t2 = _mm_shuffle_epi32(t1, 0xff);
+        t4 = _mm_slli_si128(t3, 0x4);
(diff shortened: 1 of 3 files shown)


aes/aesni/src/aes192/expand.rs
```diff
@@ -1,91 +1,107 @@
...
+                    _mm_aesimc_si128($k));
+            }
+        }
+
+        // we are being extra pedantic here to remove out-of-bound access.
+        // this should be optimized out into movups, movsd sequence
+        // note that unaligned load MUST be used here, even though we read
+        // from the array (compiler missoptimizes aligned load)
+        let (k0, k1l) = {
+            let mut t: [u8; 32] = mem::uninitialized();
```
gnzlbg: You can use `#[repr(align(16))]` here to force this array to be properly aligned, so that you can use aligned loads and stores. Something like:

```rust
#[repr(align(16))] struct u8x32([u8; 32]);
```

would do.

aes/aesni/src/aes192/expand.rs
```diff
@@ -1,91 +1,107 @@
...
+                _mm_loadu_si128(t.as_ptr() as *const __m128i),
+                _mm_loadu_si128(t.as_ptr().offset(16) as *const __m128i),
+            )
+        };
+
+        _mm_store_si128(enc_keys.as_mut_ptr(), k0);
+        _mm_store_si128(dec_keys.as_mut_ptr(), k0);
+
+        let (k1_2, k2r) = expand_round!(k0, k1l, 0x01);
+        let k1 = shuffle!(k1l, k1_2, 0);
```
gnzlbg: I am not sure what `shuffle!` does, but `k1l` is partially uninitialized, which means that `k1` might contain uninitialized bytes, so here one needs to be careful.

aes/aesni/src/aes192/expand.rs
newpavlov: Either way we'll need unaligned load here, because `key` pointer can be misaligned. I've commented it because for some strange reason I get `SIGSEGV: invalid memory reference` when I replace `_mm_loadu_si128` with `_mm_load_si128`, even though godbolt snippet (linked in the issue) results in movups for both variants.

gnzlbg:
This looks good to me. Using the `load`/`store` intrinsics doesn't look that bad at all since the code around them is already using them anyways.

aes/aesni/src/aes192/expand.rs
newpavlov: Here shuffle takes half from one variable and other half from the other, thus producing full round key.

aes/aesni/src/aes192/expand.rs
gnzlbg: Hm. `key` is just loaded by `*key`, and written to the first 24 bytes of the `[u8; 32]`. 

There are, however, two `_mm_loadu_si128` below that read from the `[u8; 32]` on the stack. These could become aligned, if the `[u8; 32]` were aligned to a 16 byte boundary. Using `#[repr(align(16))]` would allow that.

Whether this influences code generation or not, no idea. Hopefully LLVM optimizes all of this to two vector loads from `key` in which case these vector loads need to be unaligned loads.

aes/aesni/src/aes192/expand.rs
gnzlbg: Makes sense, the first half of `k1l` is properly initialized.

aes/aesni/src/aes192/expand.rs
newpavlov: I think there is a bug somewhere in the LLVM codegen, but I can't produce a minimal example to demonstrate it... Even if I'll give the whole `expand` function to godbolt it produces the same assembly for both `_mm_loadu_si128` and `_mm_load_si128`...

newpavlov: Yes it does not create temporary, as was demonstrated in the link posted in the issue and can be seen [here](https://godbolt.org/g/z4PQtN).

aes/aesni/src/aes192/expand.rs
```diff
@@ -54,7 +54,7 @@ pub(super) fn expand(key: &[u8; 24]) -> ([__m128i; 13], [__m128i; 13]) {
         // note that unaligned load MUST be used here, even though we read
         // from the array (compiler missoptimizes aligned load)
         let (k0, k1l) = {
-            let mut t: [u8; 32] = mem::uninitialized();
+            let mut t = [0u8; 32];
             ptr::write(t.as_mut_ptr() as *mut [u8; 24], *key);
```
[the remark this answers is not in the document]
newpavlov: This code worked with `uninitialized` without any problems, I've replaced it with the zeroed array because it does not change generated code, so why not remove one `uninitialized`? :)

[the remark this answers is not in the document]
gnzlbg: This is totally fine! I misread, see above!

PR #35, opened 2018-11-03
Add IDEA cipher

crodriguezvega:
Implementation of the IDEA cipher according to the [original paper](http://www.isiweb.ee.ethz.ch/papers/arch/xlai-mass-inspec-1991-2.pdf) [1] and [chapter 7](http://cacr.uwaterloo.ca/hac/about/chap7.pdf) [2] of the "Handbook of Applied Cryptography".

The values used in the test of sub key generation can be found in [2] and the test vectors used to test encryption/decryption are coming from:

https://www.cosic.esat.kuleuven.be/nessie/testvectors/bc/idea/Idea-128-64.verified.test-vectors


idea/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+license = "MIT/Apache-2.0"
+description = "IDEA block cipher"
+documentation = "https://docs.rs/idea"
+repository = "https://github.com/RustCrypto/block-ciphers"
+keywords = ["crypto", "idea", "block-cipher"]
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+block-cipher-trait = "0.5"
+opaque-debug = "0.1"
```
dignifiedquire: could you update these and match the latest definitions please?

idea/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        for i in 0..ROUNDS + 1 {
+            let j = i * 6;
+            let l = k - j;
+            
+            let mut m = 1usize;
+            let mut n = 2usize;
+            if i > 0 && i < 8 {
+                m = 2;
+                n = 1;
+            }
```
dignifiedquire: this could be written more rust like as 

```rust
let (m, n) = if i > 0 && i < 8 {
  (2, 1)
} else {
  (1, 2)
};
```

idea/Cargo.toml
crodriguezvega: Done.

idea/src/lib.rs
crodriguezvega: You're right. It's corrected now.

PR #39, opened 2018-12-16
Make blowfish generic over ByteOrder

### LINKED ISSUE #26: Endianess of Blowfish
Is there a reason for Blowfish using big-endian transformation for the integer part or would it be possible to implement a maybe? generic version of the Blowfish struct that would allow you to uses LittleEndian or BigEndian at the users will? Cause for my project I require it reading them in LittleEndian for the en/decryption to be valid.

Veykril:
This should not break the public api if I'm not mistaken due to the default type parameter but I changed the dependencies so I'm not sure if thats a breaking change in a way.

Added the typedefs since something like that was mentioned in the issue. Not sure if they really are needed though cause one could just use the type parameters without adding many more characters.

Closes #26

### DIFF
--- a/blowfish/Cargo.toml
+++ b/blowfish/Cargo.toml
@@ -10,8 +10,8 @@ keywords = ["crypto", "blowfish", "block-cipher"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
-byte-tools = "0.2"
 block-cipher-trait = "0.5"
+byteorder = { version = "1", default-features = false }
 opaque-debug = "0.1"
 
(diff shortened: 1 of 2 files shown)


newpavlov:
Thank you, looks good to me! I totally forgot that default type parameters work on stable...

`BlowfishBE` type alias is indeed unnecessary (so you can remove it), but I think `BlowfishLE` will be useful for discoverability reasons.

Veykril:
Removed it

newpavlov:
Thanks! I will publish it right away as `v0.3.1`.

PR #45, opened 2019-04-28
Chore: Update dependencies hex-literal to 0.2

tim77:


newpavlov:
Thank you, but I plan to update hex-literal version with the next minor version change which will include [MSRV](https://github.com/rust-lang/rfcs/pull/2495) change and update to 2018 edition.

PR #53, opened 2019-08-25
Formatting and minor fixes to the error types

Luro02:
According to the documentation of the Error trait, implementing is depracted and should therefore not be implemented. https://doc.rust-lang.org/std/error/trait.Error.html#method.description
I also switched to the `write!`-macro, that is used in the examples of `fmt::Display`.
https://doc.rust-lang.org/std/fmt/trait.Display.html


block-modes/src/errors.rs
```diff
@@ -11,27 +11,34 @@ pub struct BlockModeError;
 pub struct InvalidKeyIvLength;
 
 impl fmt::Display for BlockModeError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        f.write_str("BlockModeError")
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "BlockModeError")
```
newpavlov: Since we don't have any formatting arguments, I think using `f.write_str()` is perfectly fine. Otherwise with `write!` macro we rely on compiler to optimize out all unnecessary layers.

Luro02:
What is the status of this PR?

tarcieri:
Looks good to me, although I don’t have write access to this repo

tarcieri:
I'll do a `rustfmt` pass shortly here. If there are any non-`rustfmt` changes, can you please submit those separately? Thanks!

Luro02:
> can you please submit those separately?

I will definitely do this later :)

Luro02:
Well all changes were already applied, so I am deleting my fork

Issue #6, opened 2017-05-08
Develop trait for tweakable block ciphers


newpavlov:
Recently added Threefish is a [tweakable](http://people.csail.mit.edu/rivest/LiskovRivest-TweakableBlockCiphers.pdf) block cipher, thus we need an appropriate trait for it.

Currently I am thinking of using this:
```Rust
pub trait TweakableBlockCipher {
    type BlockSize: ArrayLength<u8>;
    type TweakSize: ArrayLength<u8>;

    fn encrypt(&self, block: &mut Block<Self::BlockSize>,  tweak: &Block<Self::TweakSize>);

    fn decrypt(&self, block: &mut Block<Self::BlockSize>, tweak: &Block<Self::TweakSize>);
}
```

In addition to it I think tweakable ciphers should auto implement `BlockCipher` using zero tweak.

eternaleye:
One thing worth noting is that tweak changes aren't necessarily free (some constructions even use tweak-dependent keys for the underlying block cipher), and an interface that can amortize this may be beneficial.

james-darkfox:
For HCTR the TBC trait would need to recognize Variable Input Lengths and variable tweaks (depending on the AXU it is instantiated with).

I don't think `GenericArray` can currently capture variable lengths.

newpavlov:
I will close it in favor of RustCrypto/traits#177.

PR #55, opened 2019-09-16
added HCTR

WildCryptoFox:
~~Not quite ready for publishing.~~ Ready for review!

* [x] Generic over the block cipher and AXU universal hash function.

* [x] Documentation.

* [ ] Review.

* [ ] Publish to crates.io under RustCrypto

> Why is Aes + ctr about half the performance of aesni w\ the ctr feature?

Because the latter benefits from XMM registers for the counter, while the former is more general and not currently checking if it can use XMM registers to boost aes in this case.

### DIFF
--- a/hctr/Cargo.toml
+++ b/hctr/Cargo.toml
@@ -14,6 +14,9 @@ categories = ["cryptography", "no-std"]
 name = "hctr"
 harness = false
 
+[features]
+default = ["aes", "polyval"]
+
 [dependencies.block-cipher-trait]
 version = "0.6.2"
 default-features = false
@@ -22,19 +25,25 @@ default-features = false
 version = "0.3.2"
 default-features = false
 
+[dependencies.universal-hash]
+version = "0.2.0"
+default-features = false
+
 [dependencies.aes]
 version = "0.3.2"
 default-features = false
+optional = true
 
 [dependencies.polyval]
-version = "0.0.1"
+version = "0.1.0"
+optional = true
 default-features = false
 
 [dev-dependencies.criterion]
 version = "0.3.0"
 default-features = false
 
-[dev-dependencies.aesni]
-version = "0.6.0"
-default-features = false
-features = ["ctr"]
+#[dev-dependencies.aesni]
+#version = "0.6.0"
+#default-features = false
+#features = ["ctr"]
--- a/hctr/Cargo.toml
+++ b/hctr/Cargo.toml
@@ -16,6 +16,7 @@ harness = false
 
 [features]
 default = ["aes", "polyval"]
+extended-bench = ["aez", "aesni"]
 
 [dependencies.block-cipher-trait]
 version = "0.6.2"
@@ -35,15 +36,23 @@ default-features = false
 optional = true
 
 [dependencies.polyval]
-version = "0.1.0"
+git = "https://github.com/james-darkfox/universal-hashes"
 optional = true
 default-features = false
 
-[dev-dependencies.criterion]
-version = "0.3.0"
+# extended-benchmark only. these are not used by hctr.
+[dependencies.aesni]
+version = "0.6.0"
 default-features = false
+features = ["ctr"]
+optional = true
 
-#[dev-dependencies.aesni]
-#version = "0.6.0"
-#default-features = false
-#features = ["ctr"]
+[dependencies.aez]
+git = "https://github.com/james-darkfox/aez"
+optional = true
+
+[dev-dependencies.criterion]
+git = "https://github.com/james-darkfox/criterion.rs"
+branch = "cycles"
+default-features = false
+features = ["cycles"]
--- a/hctr/Cargo.toml
+++ b/hctr/Cargo.toml
@@ -1,0 +1,40 @@
+[package]
+name = "hctr"
+version = "0.0.0"
+authors = ["RustCrypto Developers"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+description = "HCTR implementation using Aes128 and Polyval."
+documentation = "https://docs.rs/hctr"
+repository = "https://github.com/RustCrypto/block-ciphers"
+keywords = ["crypto", "hctr", "aes", "polyval", "block-cipher"]
+categories = ["cryptography", "no-std"]
+
+[[bench]]
+name = "hctr"
+harness = false
+
+[dependencies.block-cipher-trait]
+version = "0.6.2"
+default-features = false
+
+[dependencies.ctr]
+version = "0.3.2"
+default-features = false
+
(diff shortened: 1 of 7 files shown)


newpavlov:
Thank you!

I am not familiar with this construct, so I will have to read about it first to review your PR.

>Not all ADX universal hash functions are valid; there are at least two additional properties that must hold.

We probably should add marker trait(s) to describe those properties to the `universal-hash` crate. cc @tarcieri 

>Why is Aes + ctr about half the performance of aesni w\ the ctr feature?

Because `aesni` uses some platform-dependent optimizations to make code more SIMD friendly. Without it blocks have to be transferred back and forth between XMM and general purpose registers.

james-darkfox:
You're welcome!

Is there anything blocking the ctr crate from doing the same? I'm using the ctr crate at least because it is easier to reuse the same aes128 post-keying state over both the flat PRP and the CTR components to HCTR.

Aesni's Aes128Ctr takes a key and not an already instantiated Aes128. Adding that from_cipher fn would probably be sufficient, but then it'd be even more aes-specialized but I'd rather the HCTR impl be generic, not hardcoded.

james-darkfox:
Regarding the CI failure. Just missing flags for this test. `RUSTFLAGS="-Ctarget-cpu=native -Ctarget-feature=+sse2,+sse4.1,+aes,+ssse3" cargo bench`

Should the crate use cfg-targets to automatically select the aes and polyval features?

newpavlov:
>Is there anything blocking the ctr crate from doing the same?

Using the same approach may degrade performance for ciphers which do not use XMM registers for blocks processing and keep data in general purpose registers. Although it may be that bad, since those ciphers are probably will not be performant enough to begin with for several additional cycles to be noticeable (especially if users will enable SSE4.1 which has `_mm_extract_epi64`). I will need to do some experiments with it.

>Should the crate use cfg-targets to automatically select the aes and polyval features?

For now I think you can simply disable benchmarks if the required target features are not enabled.

james-darkfox:
@newpavlov The blockcipher trait could have an associated boolean constant, defaulting to false. `const USES_XMM_REGISTERS: bool = false;`

Disabling only the benchmark wouldn't be sufficient as the whole library needs to be built (or disabled). I have generalized it over BlockCipher, but until polyval implements the universal-hash trait; I can't have the library even build a generic implementation without the right features and without adding polyval's insecure-soft feature, which could be a "default" but I really do not want to enable an insecure default.

james-darkfox:
Polyval's software implementation should be constant time eventually. This PR can block on the const-time polyval.

tarcieri:
It’s already constant time. I just need to do another release. But first I wanted to add runtime feature detection.

hctr/Cargo.toml
```diff
@@ -22,19 +27,26 @@ default-features = false
...
+
 [dependencies.aes]
 version = "0.3.2"
 default-features = false
+optional = true
 
 [dependencies.polyval]
-version = "0.0.1"
+git = "https://github.com/RustCrypto/universal-hashes"
+branch = "polyval/runtime-detection"
```
james-darkfox: Blocking on https://github.com/RustCrypto/universal-hashes/pull/11

james-darkfox:
@newpavlov The [CI failure](https://travis-ci.org/RustCrypto/block-ciphers/jobs/586837672#L272) is in `lazy_static` only for the older Rust compiler. The CI is happy with my code.

The second commit generalizes over BlockCipher and UniversalHash. The API is marked as unstable until [we have a TweakableBlockCipher trait](https://github.com/RustCrypto/block-ciphers/issues/6). Once @tarcieri 's polyval PR is accepted, I'll patch the `Cargo.toml` to use the published `polyval` crate, then I'll mark this PR as "ready for review".

james-darkfox:
@tarcieri published polyval 0.1.0 which uses the `UniversalHash` trait and has no `insecure-soft` mode. @newpavlov HCTR version 0.0.0 is ready for crates.io if you're happy with it. The API will remain unstable until the `TweakableBlockCipher` trait is ready.

james-darkfox:
@newpavlov Regarding the ctr crate performance issue. What do you think about an `IntoStream` trait where `Aes128` would map directly to `Aes128Ctr`, which implements at least one of the `Stream` traits. Ideally `Aes128Ctr` can immutably refer to the `Aes128` without cloning or consuming; and the `Stream` trait could expose interleaving. I.e. `apply_keystream_and_interleave(&mut buf, |chunk| hasher.update_padded(chunk))`. What do you think?

james-darkfox:
Interleaving should be before and after each chunk. With mutable access, though HCTR only needs post-immutable interleaving.

james-darkfox:
Interleaving the second AXU with CTR doesn't appear to improve performance.

Interleaving two AXU passes (for [sphinx](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf)) benefits both small and large messages.

Interleaving all three yields worse performance than just interleaving the hashes.

For the Sphinx client use case, the benefit is to interleave the AXUs across instances.

test | hash + (ctr + hash) | hash + ctr + hash
--- | --- | ---
1K | **2.46 us** | 2.94 us
16K | 37.25 us | **37.20** us

test | hash + ctr + (hash + hash) | hash + ctr + hash + hash
--- | --- | ---
1K | **2.97 us** | 3.38 us
16K | **44.47 us** | 51.99 us

test | hash + (ctr + hash + hash) | hash + ctr + hash + hash
--- | --- | ---
1K | **3.13 us** | 3.35 us
16K | **48.34 us** | 60.82 us

james-darkfox:
hctr

james-darkfox:
@newpavlov Last commit makes this free of git dependencies. All depends are on crates.io. Just waiting on your review.

tarcieri:
@WildCryptoFox can you please rebase this? thank you!

WildCryptoFox:
@tarcieri I can do this, yes. However I'm still not sure if HCTR belongs in "block ciphers" or "aeads". It is a construction a la SIV but not restricted to AES or POLYVAL. They're merely recommended choices. A block cipher and AXU\* like POLYVAL in; wide SPRP out. Padding alone implies an MRAE.

I can adjust the implementation to use the AEAD traits and keep the layering specialization separate.

\* The AXU must also meet two properties defined in the paper for the security proofs (Section 3.3.)

tarcieri:
Yes, I'd suggest reopening this PR in the AEADs repo.

It sounds like you'd want to have several generic parameters. Several of the AEAD ciphers already do this, as it were. See [AesGcm](https://docs.rs/aes-gcm/0.5.0/aes_gcm/struct.AesGcm.html) as an example (which supports a generic nonce size).

PR #56, opened 2019-09-30
aesni -> 2018 edition

akhilles:
Requires minimum `rustc` version to be bumped to 1.31.


newpavlov:
Thank you! We plan to do a migration to 2018 edition together with new versions of the trait crates, see RustCrypto/traits#43. So I will open this PR open for the time being and will merge it right after they'll get released.

tarcieri:
Going to close this out as a bit outdated. I'll do a 2018 edition pass using `cargo fix` when we release https://github.com/RustCrypto/traits/issues/43

PR #57, opened 2019-10-12
XTS support

tomasritter:
Implemented XTS according to the specification found here:
https://web.cs.ucdavis.edu/~rogaway/papers/modes.pdf

The only difference is that this implementation can be used with ciphers different from Aes128 and Aes256. The BlockSize of that cipher must be 128bit though. 

Test without ciphertext stealing and with. Compared against output from openssl.


tomasritter:
Does it need to work with older versions of Rust or can it be merged this way?

tomasritter:
Works both for little and big endian, produces the same ciphertext as openssl

tarcieri:
Can you please rebase? Thank you!

tarcieri:
Closing as I haven't received a reply. Please reopen or resubmit if you'd like to continue this work.

PR #63, opened 2020-01-14
Add SM4 support

andelf:
Got false hope on <https://crates.io/crates/sm4>.

So I tried to write a simple implementation.


tarcieri:
@newpavlov thoughts on this?

PR #68, opened 2020-03-08
Adds benches for Serpent + questions regarding optimization

Slals:
I added benches for Serpent but results are not so satisfying, it could be x3-x10 better.

Here are the benchmark results :

```
running 2 tests
test decrypt ... bench:       8,447 ns/iter (+/- 200) = 1 MB/s
test encrypt ... bench:       8,325 ns/iter (+/- 676) = 1 MB/s
```

I have one simple question : could you guys provide some hints to improve the current Serpent implementation encrypt and decrypt speed? I don't mean Serpent specifics but more like Rust performance optimization such as `inline` and good practices.

Another question for you @tarcieri, earlier you told me about checking `branch instruction` on ASM, since I'm kinda newbie on this could you tell me what I have to consider regarding this?

Thanks!

### DIFF
--- a/serpent/src/lib.rs
+++ b/serpent/src/lib.rs
@@ -28,9 +28,7 @@ pub struct Serpent {
     k: Subkeys,
 }
 
-fn get_bit(x: usize, i: usize) -> u8 {
-    (x >> i) as u8 & 0x01
-}
+fn get_bit(x: usize, i: usize) -> u8 { (x >> i) as u8 & 0x01 }
 
 fn linear_transform_bitslice(input: Block128, output: &mut Block128) {
     let mut words = [0u32; 4];
@@ -71,7 +69,8 @@ fn round_bitslice(
     b_i: Block128,
     k: Subkeys,
     b_output: &mut Block128,
-) {
+)
+{
     let xored_block = xor_block(b_i, k[i]);
 
     let s_i = apply_s_bitslice(i, xored_block);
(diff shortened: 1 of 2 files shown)


tarcieri:
@Slals I don't know much about Serpent, however those numbers certainly do seem quite slow

serpent/src/lib.rs
```diff
@@ -229,6 +227,7 @@ impl BlockCipher for Serpent {
         Self::new_varkey(key).unwrap()
     }
 
+    #[inline(always)]
```
tarcieri: Did these actually help? IMO every `#[inline]` directive should have a comment explaining what it's doing and why that's better than what the LLVM optimizer did by default

jack-signal:
Right now it seems like Serpent is implemented in this crate using table lookups instead of the usual bitsliced sbox expressions. Changing that would probably provide about a 50x-100x performance improvement. In bitsliced mode it is also easy to use SIMD for parallel execution.

tarcieri:
@Slals would you mind changing this PR to just include the benchmarks for now? Thanks!

Slals:
Thanks @jack-signal for your feedback. I did't take the time to make the change in order to speed it. 

@tarcieri I removed the "inlines" tags if it's what you needed.

tarcieri:
@Slals there are some conflicting changes in `serpent/src/lib.rs`. They might just go away if you run `rustfmt` on this branch. Otherwise you need to rebase.

tarcieri:
I think the original goal of this PR was actually accomplished in #105 last May. Going to close this out.

PR #100, opened 2020-05-27
twofish: 2018 edition and `block-cipher` crate upgrade

tarcieri:
Upgrades to Rust 2018 edition and the (now 2018 edition) `block-cipher` crate (formerly `block-cipher-trait`).


Cargo.lock
```diff
@@ -42,17 +42,7 @@ version = "0.7.0-pre"
 source = "git+https://github.com/RustCrypto/traits.git#9cfac08fc7c151122ddebd580e0754c240d23db2"
 dependencies = [
  "blobby",
- "generic-array 0.14.1",
-]
-
-[[package]]
-name = "block-cipher-trait"
```
tarcieri, line 49: 🎉 last one!

PR #101, opened 2020-05-27
Bump all versions to `*-pre`

tarcieri:
...to reflect the breaking 2018 edition and `block-cipher` updates


newpavlov:
When you'll start publishing please omit `kuznyechik` and `magma`. To be compatible with common implementations BE order has to be used, instead of the wrongly used LE, so I would like to fix it before the breaking update.

PR #109, opened 2020-05-27
aesni: use MaybeUninit::zeroed() to init XMM registers

tarcieri:
Resolves clippy lints about soundness, and will be optimized away by the compiler:

https://rust.godbolt.org/z/inEYXR


newpavlov:
Why not use shorter `mem::zeroed()` instead? AFAIK it's not getting deprecated as `mem::uninitialied()`

tarcieri:
Aah, well then...

tarcieri:
Opened #110

PR #113, opened 2020-05-30
rustfmt: use stock configuration

tarcieri:
This prevents linewrapping in a large number of cases, arguably improving readability.


tarcieri:
All of the other repos are using the stock rustfmt configuration, so if nothing else this is more consistent.

PR #116, opened 2020-06-03
Fix CI config

newpavlov:


.github/workflows/aesni.yml
```diff
@@ -15,6 +15,8 @@ defaults:
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
+  RUSTDOCFLAGS: "-C target-feature=+aes,+ssse3"
+  RUSTFLAGS: "-C target-feature=+aes,+ssse3"
```
tarcieri, line 19: It looks like `RUSTFLAGS` is duplicated here? I'm not sure how GitHub actions handles that

PR #118, opened 2020-06-03
Fix GOST byte order

newpavlov:


magma/tests/lib.rs
```diff
@@ -23,3 +30,30 @@ fn magma_test() {
...
+    let plaintext = b"The quick brown fox jumps over the lazy dog\n";
+
+    let c = Cipher::new_var(&key, &iv).unwrap();
+    let ct = c.encrypt_vec(plaintext);
+    for b in ct {
+        print!("0x{:02X}, ", b);
+    }
+    println!();
+}
+*/
```
tarcieri, line 59: Strange, it seems rustfmt just wants to delete this entire block.

Perhaps use line comments instead of block comments? Or I wonder if there's an option to shut it off...

tarcieri: Here's an issue about it...

https://github.com/rust-lang/rustfmt/issues/4028

newpavlov: Yeah, certainly an annoying bug...

PR #120, opened 2020-06-03
magma: fix rustfmt

tarcieri:
For whatever reason rustfmt wants to aggressively strip trailing comments.

This adds a small hack to preserve the commented-out code.

### DIFF
--- a/magma/tests/lib.rs
+++ b/magma/tests/lib.rs
@@ -1,5 +1,5 @@
 #![no_std]
-#![cfg_attr(rustfmt, rustfmt_skip)]
+
 use magma;
 
 use block_cipher::generic_array::GenericArray;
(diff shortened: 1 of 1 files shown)


magma/tests/lib.rs
```diff
@@ -1,5 +1,5 @@
 #![no_std]
-#![cfg_attr(rustfmt, rustfmt_skip)]
```
tarcieri, line 2: This appears to have no effect.

See also: https://github.com/rust-lang/rustfmt/issues/4028

newpavlov: It does disable rustfmt in the test function. I would prefer to keep data there 16 bytes per line.

tarcieri: Ok. I'd suggest using the new nested attribute syntax:

```rust
#![rustfmt::skip]
```

tarcieri: Oh whoops, that isn't stable for inner attributes yet :cry:

newpavlov: Looks like this feature is still unstable.

tarcieri:
If you'd prefer it, that's fine with me

PR #125, opened 2020-06-05
aes-soft v0.4.0

tarcieri:
### Changed
- Bump `block-cipher` dependency to v0.7 ([#86], [#122])
- Update to Rust 2018 edition ([#86])
 
[#122]: https://github.com/RustCrypto/block-ciphers/pull/122
[#86]: https://github.com/RustCrypto/block-ciphers/pull/86

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
 
 [[package]]
 name = "aes-soft"
-version = "0.4.0-pre"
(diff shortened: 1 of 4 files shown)


newpavlov:
Maybe update all versions as part of one PR? Especially since CI runs all jobs either way.

tarcieri:
@newpavlov sure, will do that...

PR #126, opened 2020-06-05
aes: release new crates

tarcieri:
Releases the following versions of the `aes/` crates:

- `aes` v0.4.0
- `aes-soft` v0.4.0
- `aesni` v0.7.0


tarcieri:
@newpavlov going to start with these. They can unblock getting some of the other repos updated.

PR #134, opened 2020-06-19
Implement modes defined in GOST R 34.13-2015

newpavlov:
The GOST modes can be viewed as a generalization of OFB, CBC and CFB modes, which break some assumptions which were used in `BlockMode` and stream cipher traits.


newpavlov:
@tarcieri 
I must say trait bounds for the new types are something. :)

Would you like to review the changes before merge? Also note that this PR also introduces a breaking change to `block-modes` by adding `IvSize` associated type to the `BlockMode` trait.

tarcieri:
> I must say trait bounds for the new types are something. :)

Const generics can't come soon enough

> Would you like to review the changes before merge? 

Just glanced over it and it looks ok to me

tarcieri:
> I must say trait bounds for the new types are something. :)

Const generics can't come soon enough.

tarcieri:
Whoops, fat fingered the close button 😉

newpavlov:
>Const generics can't come soon enough

I worry that the planned initial version of const generics will not be powerful enough to encode the bounds used in this PR...

Issue #25, opened 2018-08-14
AES implementation should be chosen at runtime rather than compile time


upsuper:
AES implementation should be chosen at runtime rather than compile time, otherwise it makes people very hard to ship products with this crate, because they cannot choose the environment a product would be run on.

newpavlov:
Switch between implementations should be done as high-level as possible, not at the block cipher level. There is a [`nocapture`](https://github.com/RustCrypto/block-ciphers/pull/22) feature for this use-case, but you will need to be sure to enable necessary target features for proper inlining.

newpavlov:
I believe ideally we need a support for this at the language level. I've described my view [here](https://internals.rust-lang.org/t/pre-pre-rfc-target-restriction-contexts/7163).

tarcieri:
What I'd like in the short term:

- On platforms that have no `aesni` impl, use the software fallback
- On platforms that do (presently `x86`/`x86_64`) require the appropriate target features are enabled (i.e. `+aes,+ssse3`) and refuse to compile if they are not. Use runtime detection to determine if AES-NI is available, and if not, fall back to the software implementation

kazcw:
> On platforms that do (presently x86/x86_64) require the appropriate target features are enabled (i.e. +aes,+ssse3) and refuse to compile if they are not.

Why not `[target_feature(enable = "aes")]`?

tarcieri:
> Why not `[target_feature(enable = "aes")]`

Because it silently falls back to only using the software implementation of AES. In fact, if you compiled with `RUSTFLAGS=-Ctarget-feature=+aes` today, that is exactly what will happen, because using `aesni` requires `+aes,+ssse3`, and `+aes` alone will silently fall back.

I have plastered Miscreant with instructions to use `RUSTFLAGS` or `~/.cargo/config` to enable the appropriate target features, but even then they changed. In absence of that, I am pretty sure nearly everyone using the `aes` crate is using `aes-soft` instead of `aesni`.

[It would be nice if the ergonomics of target features were better](https://internals.rust-lang.org/t/pre-pre-rfc-target-restriction-contexts/7163/4), but failing that, if the `aes` crate silently falls back to `aes-soft` with no runtime AES-NI detection unless the target features are correctly configured, then nobody is going to bother to configure their builds correctly, and everything will use the software implementation. I would consider that a bad outcome, as runtime AES-NI detection is the norm for crypto libraries outside the Rust ecosystem.

[...]

PR #135, opened 2020-06-25
Towards AES runtime detection

newpavlov:
This is a preliminary work towards solving #25.

For now I've simply removed SIMD parts from `aes-soft`. It reduced state sizes to the level of `aesni`, but as expected also degraded performance of parallel block processing (on my PC for `Aes128` from 70 MB/s to 21 MB/s). I think it should be possible to keep SIMD paths without keeping the whole expanded state.

I think I will move towards merging `aesni`, `aes-soft` and `aes-ctr` into `aes` crate.


tarcieri:
> I think I will move towards merging `aesni`, `aes-soft` and `aes-ctr` into `aes` crate.
💯

mat-gas:
Any news on this PR? some much looking forward to having it implemented :)

tarcieri:
This PR has been obsoleted by the move to fixsliced AES in the `aes-soft` crate

PR #165, opened 2020-10-13
aesni: Add target_feature annotations to allow intrinsic inlining

jack-signal:
When built with nocheck feature, the AES-NI intrinsics are not inlined causing a severe performance drop. (https://github.com/rust-lang/rust/issues/53069)

When compiled with `--features=nocheck`, improvement is notable (all numbers MB/s as reported by `cargo bench` on an i7-10510U, rustc 1.47)

| Benchmark | Current master (+aes) | Current master (nocheck) | This PR (nocheck) |
-----------------| ---------------------| --------------------| --------------------
| aes128_encrypt | 615 | 285 | 516 |
| aes128_encrypt8 | 2782 | 551 | 2327 |
| aes192_encrypt | 533 | 242 | 444 |
| aes192_encrypt8 | 2327 | 462 | 2000 |
| aes256_encrypt | 457 | 210 | 400 |
| aes256_encrypt8 | 2000 | 393 | 1753 |

When compiled with `RUSTFLAGS="-C target-feature=+aes,+ssse3"` there was no performance change (based on `cargo bench` output)

Decryption could be updated similarly. I didn't need it, but could update that as well if desired.

### DIFF
--- a/aes/aesni/src/aes128.rs
+++ b/aes/aesni/src/aes128.rs
@@ -92,11 +92,11 @@ impl BlockCipher for Aes128 {
 
     #[inline]
     fn decrypt_block(&self, block: &mut Block128) {
-        let keys = self.decrypt_keys;
-
-        // Safety: `loadu` and `storeu` support unaligned access
-        #[allow(clippy::cast_ptr_alignment)]
-        unsafe {
+        #[inline]
+        #[target_feature(enable = "aes")]
+        unsafe fn aes128_decrypt1(block: &mut Block128, keys: &[__m128i; 11]) {
+            // Safety: `loadu` and `storeu` support unaligned access
+            #[allow(clippy::cast_ptr_alignment)]
             let mut b = _mm_loadu_si128(block.as_ptr() as *const __m128i);
             b = _mm_xor_si128(b, keys[10]);
             b = _mm_aesdec_si128(b, keys[9]);
@@ -111,6 +111,8 @@ impl BlockCipher for Aes128 {
             b = _mm_aesdeclast_si128(b, keys[0]);
             _mm_storeu_si128(block.as_mut_ptr() as *mut __m128i, b);
         }
+
+        unsafe { aes128_decrypt1(block, &self.decrypt_keys) }
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


tarcieri (approved):
This looks good to me

> Decryption could be updated similarly. I didn't need it, but could update that as well if desired.

That'd be appreciated!

Will wait to hear from @newpavlov for a bit before merging.

jack-signal:
Added decryption in 82970df, improvement looks to be similar.

tarcieri:
@jack-signal thanks!

clippy error looks unrelated. I can try to fix that up in a separate PR.

newpavlov:
Strictly speaking doing so is incorrect and even having the `nocheck` feature in the first place. The methods would have to be marked as `unsafe` since we do not check target feature availability neither at compile, nor run time. The feature was added mostly to make easier generation of docs. Do you plan to use it in practice?

jack-signal:
> Do you plan to use it in practice with a runtime detection on top 

Yes in my code I have an enum 

```
pub enum Aes256 {
    Soft(Box<aes_soft::Aes256>),
    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
    AesNi(Box<aesni::Aes256>),
    #[cfg(target_arch = "aarch64")]
    Aarch64(Box<aarch64::Aes256Aarch64>),
}
```

and then `Aes256::new` chooses which to use based on runtime checks (`is_x86_feature_detected`, or `getauxval` for Android/Linux on Aarch64)

> The methods would have to be marked as unsafe since we do not check target feature availability neither at compile, nor run time. 

But that's true for `nocheck` with or without this change, right? I thought the whole point of this feature was to allow for higher levels to choose which implementation to use.

newpavlov:
>But that's true for nocheck with or without this change, right?

Yes. Unfortunately it is not possible to mark safe trait methods as `unsafe` in implementation.

I guess we can merge it as a temporary solution until #135 lands.

aes/aesni/src/aes128.rs
```diff
@@ -19,8 +19,9 @@ pub struct Aes128 {
 impl Aes128 {
     #[inline(always)]
     pub(crate) fn encrypt8(&self, mut blocks: [__m128i; 8]) -> [__m128i; 8] {
-        let keys = self.encrypt_keys;
-        unsafe {
+        #[inline]
+        #[target_feature(enable = "aes")]
```
newpavlov, line 23: Please add a full list of features, which you can find in `src/target_checks.rs`. It should improve performance a bit further.

newpavlov (commented):
Also add similar attributes for the CTR code?

aes/aesni/src/aes128.rs
jack-signal: Per https://doc.rust-lang.org/reference/attributes/codegen.html "aes" implies "sse2", and the only use of SSSE3 is in counter mode for pshufb. So I'm not sure if it makes any difference, though I could change it if you prefer for consistency.

newpavlov: >"aes" implies "sse2"

Ah, good to know.

PR #172, opened 2020-10-23
[WIP] Prototype of bitsliced AES using Bs8State<u32>

peterdettman:
Passes the aes-soft tests and gives better benchmarks on my haswell laptop than the 8-block `Bs8State<u32x4>` version, although I'm not sure if I have SIMD properly configured there (I at least have avx on).

As written, this replaces the `Bs8State<u32x4>` version, retaining the 8-block parallel API, but it's natural parallelism is just 2 blocks. Ideally it would be a configurable option but I'm not sure how to go about setting that up.

### DIFF
--- a/aes/aes-soft/src/bitslice.rs
+++ b/aes/aes-soft/src/bitslice.rs
@@ -4,8 +4,8 @@
     clippy::unreadable_literal
 )]
 
-use crate::consts::U32X4_1;
-use crate::simd::u32x4;
+//use crate::consts::U32X4_1;
+//use crate::simd::u32x4;
 use byteorder::{ByteOrder, LE};
 use core::ops::{BitAnd, BitXor, Not};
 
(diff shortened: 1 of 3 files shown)


tarcieri:
> As written, this replaces the Bs8State<u32x4> version, retaining the 8-block parallel API, but it's natural parallelism is just 2 blocks.

This would better align with the fixsliced implementations which also have a 2-block parallelism.

If you're interested in continuing this, can you rebase?

Changing `ParBlocks` would be a SemVer breaking change, but it's one I think we should consider to align the implementations.

Some of the consumers of this crate are hardcoded to assume `ParBlocks = U8`, but we should really fix that so consumers are generic over the number of parallel blocks everywhere.

tarcieri:
@peterdettman alternatively, we could explore implementing fixsliced decryption. It would need the reverse S-box which works similarly to the current implementation.

peterdettman:
Closing in favour of focusing on fixsliced implementation (#179).

PR #173, opened 2020-10-23
[WIP] Prototype of bitsliced AES using Bs8State<u64>

peterdettman:
Passes the aes-soft tests and gives much better benchmarks on my haswell laptop than the 8-block `Bs8State<u32x4>` version (twice as fast as https://github.com/RustCrypto/block-ciphers/pull/172), although I'm not sure if I have SIMD properly configured there (I at least have avx on).

As written, this replaces the `Bs8State<u32x4>` version, retaining the 8-block parallel API, but it's natural parallelism is just 4 blocks. Ideally it would be a configurable option but I'm not sure how to go about setting that up.

### DIFF
--- a/aes/aes-soft/src/bitslice.rs
+++ b/aes/aes-soft/src/bitslice.rs
@@ -4,8 +4,8 @@
     clippy::unreadable_literal
 )]
 
-use crate::consts::U32X4_1;
-use crate::simd::u32x4;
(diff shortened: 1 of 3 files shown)


tarcieri:
Can you rebase? See also the comments I left on #172

peterdettman:
Closing in favour of focusing on fixsliced implementation (#180).

PR #174, opened 2020-10-23
aes-soft: fixslicing

### LINKED ISSUE #162: aes: implement fixslicing? (portable and ASM versions)
A newly released paper describes how to apply a bitslicing-like technique called "fixslicing" to AES:

https://eprint.iacr.org/2020/1123

> Fixslicing AES-like Ciphers: New bitsliced AES speed records on ARM-Cortex M and RISC-V
>
> The fixslicing implementation strategy was originally introduced as a new representation for the hardware-oriented GIFT block cipher to achieve very efficient software constant-time implementations. In this article, we show that the fundamental idea underlying the fixslicing technique is not of interest only for GIFT, but can be applied to other ciphers as well. Especially, we study the benefits of fixslicing in the case of AES and show that it allows to reduce by 41% the amount of operations required by the linear layer when compared to the current fastest bitsliced implementation on 32-bit platforms. Overall, we report that fixsliced AES-128 allows to reach 83 and 99 cycles per byte on ARM Cortex-M and RISC-V respectively (assuming pre-computed round keys), improving the previous records on those platforms by 17% and 20%. In order to highlight that our work also directly improves masked implementations that rely on bitslicing, we report implementation results when integrating first-order masking that outperform by 12% the fastest results reported in the literature on ARM Cortex-M4. Finally, we demonstrate the genericity of the fixslicing technique for AES-like designs by applying it to the Skinny-128 tweakable block ciphers.

As part of the paper they've also posted portable C implementations we might consider translating:

https://github.com/aadomn/aes/tree/master/opt32/fixslicing

...along with optimized Cortex-M assembly we might consider integrating (optionally, behind a feature flag, possibly using the new inline ASM syntax):

https://github.com/aadomn/aes/tree/master/armcortexm/fixslicing

tarcieri:
Closes #162

Translates the portable C implementation of the "fixslicing" technique described in:

https://eprint.iacr.org/2020/1123.pdf

Note that the reference implementation only supports AES-128/AES-256 *encryption*.

### DIFF
--- a/aes/aes-soft/src/fixslice.rs
+++ b/aes/aes-soft/src/fixslice.rs
@@ -1,0 +1,893 @@
+//! Fixsliced implementations of AES-128 and AES-256 (encryption-only)
+//! adapted from the C implementation.
+//!
+//! All implementations are fully bitsliced and do not rely on any
+//! Look-Up Table (LUT).
+//!
+//! See the paper at <https://eprint.iacr.org/2020/1123.pdf> for more details.
+//!
+//! # Author (original C code)
+//!
+//! Alexandre Adomnicai, Nanyang Technological University, Singapore
+//! <alexandre.adomnicai@ntu.edu.sg>
+//!
+//! Originally licensed MIT. Relicensed as Apache 2.0+MIT with permission.
+
+use cipher::{
+    block::Key,
+    consts::{U16, U2, U32},
+    generic_array::GenericArray,
+    BlockCipher, NewBlockCipher,
+};
+use core::convert::TryInto;
+
+/// 128-bit AES block
+// TODO(tarcieri): unify with other `Block` type aliases
+type Block = GenericArray<u8, U16>;
+
+/// 2x128-bit AES blocks
+// TODO(tarcieri): unify with other `ParBlocks` type aliases
+type ParBlocks = GenericArray<Block, U2>;
+
+/// AES-128 round keys
+type RKeys128 = [u32; 88];
+
+/// AES-256 round keys
+type RKeys256 = [u32; 120];
+
+/// 256-bit internal state
+type State = [u32; 8];
+
+/// AES-128 with a fully fixsliced implementation
+#[derive(Clone)]
+pub struct Aes128Fixsliced {
+    /// Round keys
+    rkeys: RKeys128,
+}
+
+impl NewBlockCipher for Aes128Fixsliced {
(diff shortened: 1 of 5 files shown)


tarcieri:
It's failing to pass the test vectors, unfortunately.

If I can figure out what's wrong, it seems significantly faster than the current implementation (even using 32-bit operations on a 64-bit Intel CPU):

```
test aes128_encrypt           ... bench:         668 ns/iter (+/- 54) = 23 MB/s
test aes128_encrypt8          ... bench:       3,114 ns/iter (+/- 530) = 41 MB/s
test aes128_encrypt2_fixsliced ... bench:         290 ns/iter (+/- 38) = 110 MB/s
```

```
test aes256_encrypt           ... bench:         949 ns/iter (+/- 105) = 16 MB/s
test aes256_encrypt8          ... bench:       4,332 ns/iter (+/- 412) = 29 MB/s
test aes256_encrypt_fixsliced ... bench:         409 ns/iter (+/- 167) = 78 MB/s
```

aes/aes-soft/src/fixslice.rs
```diff
@@ -0,0 +1,184 @@
...
+//! See the paper at https://eprint.iacr.org/2020/1123.pdf for more details.
+//!
+//! # Author (original C code)
+//!
+//! Alexandre Adomnicai, Nanyang Technological University, Singapore
+//! <alexandre.adomnicai@ntu.edu.sg>
+
+// MIT License
+//
+// Copyright (c) 2020 Alexandre Adomnicai
```
tarcieri, line 16: Note: I asked for permission to dual license the translation as Apache2/MIT and received it

tarcieri:
The tests now pass, and I was able to completely eliminate any usages of unsafe code 🎉 

I think we should definitely figure out how to integrate this, but the next steps are somewhat tricky due to what this code lacks:

- AES-192 is not supported
- Decryption is not supported

So we need to keep the existing bitsliced implementation around for both of these cases, and figure out some way to merge them together.

tarcieri:
I think this is ready to review/merge with the caveat that the goal is to further unify the implementation with the existing bitsliced one.

I can open an issue with a plan for how to do that, or maybe just open some follow-up PRs.

tarcieri:
Going to go ahead and merge this, then investigate ways of unifying the implementations in a follow-up PR.

PR #175, opened 2020-10-24
Perf. opt. for 1-block (un)bitslice

peterdettman:
Uses a more efficient algorithm for:
- bit_slice_4x4_with_u16
- un_bit_slice_4x4_with_u16


tarcieri:
@peterdettman FYI, you might also take a look at #174

tarcieri (approved):

PR #176, opened 2020-10-25
aes-soft: use fixslicing for AES encryption

tarcieri:
Continuation of the work from #174.

Fixslicing is defined for ~~AES-128/AES-256~~ (now with AES-192!) encryption only, but we can replace bitslicing in these capacities with the faster fixslicing approach.

This is useful for AES-CTR, which needs only the encryption operation.

Performance is improved by ~3X as measured on an Intel Core i9 (despite the fixslicing implementation being 32-bit only)

cc @peterdettman

# Benchmarks

## Before

```
test aes128_decrypt  ... bench:         619 ns/iter (+/- 53) = 25 MB/s
test aes128_decrypt8 ... bench:       3,313 ns/iter (+/- 602) = 38 MB/s
test aes128_encrypt  ... bench:         608 ns/iter (+/- 80) = 26 MB/s
test aes128_encrypt8 ... bench:       3,064 ns/iter (+/- 227) = 41 MB/s

test aes256_decrypt  ... bench:         878 ns/iter (+/- 114) = 18 MB/s
test aes256_decrypt8 ... bench:       4,524 ns/iter (+/- 269) = 28 MB/s
test aes256_encrypt  ... bench:         864 ns/iter (+/- 49) = 18 MB/s
test aes256_encrypt8 ... bench:       4,274 ns/iter (+/- 296) = 29 MB/s
```

## After

```
test aes128_decrypt  ... bench:         570 ns/iter (+/- 65) = 28 MB/s
test aes128_decrypt8 ... bench:       2,904 ns/iter (+/- 123) = 44 MB/s
test aes128_encrypt  ... bench:         271 ns/iter (+/- 51) = 59 MB/s
test aes128_encrypt8 ... bench:       1,060 ns/iter (+/- 45) = 120 MB/s

test aes256_decrypt  ... bench:         840 ns/iter (+/- 35) = 19 MB/s
test aes256_decrypt8 ... bench:       4,194 ns/iter (+/- 242) = 30 MB/s
test aes256_encrypt  ... bench:         374 ns/iter (+/- 22) = 42 MB/s
test aes256_encrypt8 ... bench:       1,533 ns/iter (+/- 100) = 83 MB/s
```

(AES-192 is unchanged, as it still uses the original bitsliced implementation)

PR #177, opened 2020-10-26
aes-soft: fixslicing for AES-192 encryption

peterdettman:
@tarcieri I figured it's easier to PR this against the current master and let you integrate it into #176 .


tarcieri:
Nice, thanks!

PR #180, opened 2020-10-27
Add  64-bit fixslice AES implementation

peterdettman:
- protoype: replaces 32-bit in lib.rs, impls.rs

Not familiar enough with rust to attempt configuration settings.

### DIFF
--- a/aes/aes-soft/src/impls.rs
+++ b/aes/aes-soft/src/impls.rs
@@ -13,7 +13,7 @@ use crate::{
     },
     consts::U32X4_0,
     expand::expand_key,
-    fixslice::{self, FixsliceKeys128, FixsliceKeys192, FixsliceKeys256},
+    fixslice::{self, FixsliceKeys128, FixsliceKeys192, FixsliceKeys256, FIXSLICE_BLOCKS},
     simd::u32x4,
     Block, ParBlocks,
 };
@@ -70,7 +70,7 @@ macro_rules! define_aes_impl {
 
             #[inline]
             fn encrypt_block(&self, block: &mut Block) {
-	            let mut blocks = [Block::default(); 2];
+                let mut blocks = [Block::default(); FIXSLICE_BLOCKS];
                 blocks[0].copy_from_slice(block);
                 $fixslice_encrypt(&self.enc_keys, &mut blocks);
                 block.copy_from_slice(&blocks[0]);
@@ -85,7 +85,7 @@ macro_rules! define_aes_impl {
 
             #[inline]
             fn encrypt_blocks(&self, blocks: &mut ParBlocks) {
-                for chunk in blocks.chunks_mut(2) {
+                for chunk in blocks.chunks_mut(FIXSLICE_BLOCKS) {
                     $fixslice_encrypt(&self.enc_keys, chunk);
                 }
             }
--- a/aes/aes-soft/src/fixslice32.rs
+++ b/aes/aes-soft/src/fixslice32.rs
@@ -1,0 +1,1068 @@
+//! Fixsliced implementations of AES-128, AES-192 and AES-256 (encryption-only)
+//! adapted from the C implementation.
+//!
+//! All implementations are fully bitsliced and do not rely on any
+//! Look-Up Table (LUT).
+//!
+//! See the paper at <https://eprint.iacr.org/2020/1123.pdf> for more details.
+//!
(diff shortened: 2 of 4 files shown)


tarcieri:
@peterdettman it'd be good to preserve both the 32-bit and 64-bit backends, but select which one to use based on the host CPU. You can do something like:

```
 #[cfg(target_pointer_width = "32")] 
mod fixslice32;
 #[cfg(target_pointer_width = "64")] 
mod fixslice64;
```

If you're having trouble, the `cfg-if` crate can help:

https://github.com/alexcrichton/cfg-if

tarcieri:
Benchmarks look very nice:

```
test aes128_encrypt8 ... bench:         626 ns/iter (+/- 81) = 204 MB/s
test aes192_encrypt8 ... bench:         742 ns/iter (+/- 76) = 172 MB/s
test aes256_encrypt8 ... bench:         851 ns/iter (+/- 162) = 150 MB/s
```

peterdettman:
@tarcieri Yeah, I certainly want both to coexist, but I'd really prefer to leave the cfg stuff to you. It's not just cfg either; the 64-bit version does 4 blocks at a time (instead of 2) and somehow the encrypt_block(s) methods in impl.rs will need to know how many blocks the inner implementation is handling.

tarcieri:
@peterdettman okay, if you'd like I can try to handle the gating.

How about renaming the existing implementation to `fixslice32.rs` and then removing the `mod fixslice32` directive for it, and I can do an additional pass?

> It's not just cfg either; the 64-bit version does 4 blocks at a time (instead of 2) and somehow the encrypt_block(s) methods in impl.rs will need to know how many blocks the inner implementation is handling.

Can you make them all 8 for now and add a TODO to change `ParBlocks` in the next release? It's a breaking change.

I think there's also some discussion to be had there. Perhaps we should try to align all of the implementations exposed via the `aes` crate to be 4-blocks (which is also the ideal number for AES-NI on modern Intel archs).

peterdettman:
OK, I've renamed fixslice.rs to fixslice32.rs and lib.rs now switches b/w them on `cfg(target_pointer_width = "64")`. The fixslice implementations now have a FIXSLICE_BLOCKS constant that impls.rs uses to break up the 8 blocks into 2 or 4 chunks respectively.

aes/aes-soft/src/impls.rs
```diff
@@ -24,6 +24,7 @@ macro_rules! define_aes_impl {
         $key_size:ty,
         $rounds:expr,
         $rounds2:ty,
+        $fixslice_blocks:expr,
```
tarcieri, line 27: Since this is the same for all impls it doesn't need to be a macro parameter

aes/aes-soft/src/impls.rs
```diff
@@ -70,7 +71,7 @@ macro_rules! define_aes_impl {
 
             #[inline]
             fn encrypt_block(&self, block: &mut Block) {
-	            let mut blocks = [Block::default(); 2];
+                let mut blocks = [Block::default(); $fixslice_blocks];
```
tarcieri, line 74: You can use the `FIXSLICE_BLOCKS` constant directly here:

```suggestion
                let mut blocks = [Block::default(); FIXSLICE_BLOCKS];
```

aes/aes-soft/src/impls.rs
```diff
@@ -85,7 +86,7 @@ macro_rules! define_aes_impl {
 
             #[inline]
             fn encrypt_blocks(&self, blocks: &mut ParBlocks) {
-                for chunk in blocks.chunks_mut(2) {
+                for chunk in blocks.chunks_mut($fixslice_blocks) {
```
tarcieri, line 89: ...also here:

```suggestion
                for chunk in blocks.chunks_mut(FIXSLICE_BLOCKS) {
```

tarcieri:
@peterdettman looks great aside from one nit!

peterdettman:
Fixed nits and rebased.

peterdettman:
@tarcieri I’d be interested in seeing some arm (32/64) benchmarks for these fixslice implementations.

tarcieri:
@peterdettman I'll be looking into benchmarking it on Cortex-A and Cortex-M (32-bit).

Somewhere around here I have some `aarch64` devices (RasPi) I could benchmark it on too.

PR #184, opened 2020-10-28
Improved AES fixslice MixColumns algorithm(s)

peterdettman:
Not just reformatting! Improved formula for the several MixColumn variants needed by the fixslice AES with fewer operations and less data dependencies.

@aadomn Although I've done these somewhat intuitively, I believe they can be derived (e.g. for the first-round MixColumns) by grouping the initial equations of [Figure 6](https://eprint.iacr.org/2020/1123.pdf) using:

`rot_cols_2(rot_rows_2(x)) ^ rot_cols_1(rot_rows_3(x)) == rot_cols_2(rot_rows_2(x ^ rot_cols_3(rot_rows_1(x))))`

EDIT: Corrected equation.

### DIFF
--- a/aes/aes-soft/src/fixslice32.rs
+++ b/aes/aes-soft/src/fixslice32.rs
@@ -587,184 +587,113 @@ fn sbox_nots(state: &mut [u32]) {
 
 /// Computation of the MixColumns transformation in the fixsliced representation
 /// used for rounds i s.t. (i%4) == 0.
+#[rustfmt::skip]
 fn mixcolumns_0(state: &mut State) {
-    let mut tmp3 = rotate_rows_1(rotate_columns_3(state[7]));
-    let tmp0 = state[7] ^ tmp3;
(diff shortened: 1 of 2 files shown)


aadomn:
@peterdettman I'm not sure to understand how it speeds up the calculations? According to your equations it requires the same number of `rot_cols` and `rot_rows`, but maybe I'm missing something?

peterdettman:
@aadomn Check the code, after grouping that way there are more common terms.

peterdettman:
@aadomn e.g. in your paper you give the cost of Figure 6 as:

> 27 XOR, 48 AND and 24 OR instructions on top of 16 circular and 48 logical shifts

but `mixcolumns_0` in this PR uses only 27 XOR, 32 AND and 16 OR, plus 32 circular shifts.

aadomn:
@peterdettman Ok I got it! Thanks a lot for sharing this I'll mention it in the paper

PR #192, opened 2020-10-29
Comprehensive refactoring of fixslice code

peterdettman:
- Rename methods for consistency with AES specification
- De-unroll several loops
- Macro-ize the (inv_)mix_columns definitions

@tarcieri Let me know if you see anything else annoying and I can add it here too.


tarcieri:
Nice, I was also thinking of similar ways to DRY out the more repetitive parts of the code when I was originally translating it.

PR #195, opened 2020-11-02
Implement alternative semi-fixslice

peterdettman:
- under cfg feature 'semi_fixslice'
- reduces code size at small cost to performance

Alternative, slightly simpler variant described in [the paper](https://eprint.iacr.org/2020/1123). I'm not sure if this is the ideal way to present the alternative, but it gets working code in place.

### DIFF
--- a/aes/aes-soft/src/fixslice32.rs
+++ b/aes/aes-soft/src/fixslice32.rs
@@ -62,12 +62,21 @@ pub(crate) fn aes128_key_schedule(key: &GenericArray<u8, U16>) -> FixsliceKeys12
     }
 
     // Adjust to match fixslicing format
-    for i in (8..72).step_by(32) {
-        inv_shift_rows_1(&mut rkeys[i..(i + 8)]);
-        inv_shift_rows_2(&mut rkeys[(i + 8)..(i + 16)]);
-        inv_shift_rows_3(&mut rkeys[(i + 16)..(i + 24)]);
+    #[cfg(feature = "semi_fixslice")]
+    {
+        for i in (8..88).step_by(16) {
+            inv_shift_rows_1(&mut rkeys[i..(i + 8)]);
+        }
     }
-    inv_shift_rows_1(&mut rkeys[72..80]);
+    #[cfg(not(feature = "semi_fixslice"))]
+    {
+        for i in (8..72).step_by(32) {
+            inv_shift_rows_1(&mut rkeys[i..(i + 8)]);
+            inv_shift_rows_2(&mut rkeys[(i + 8)..(i + 16)]);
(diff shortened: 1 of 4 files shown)


tarcieri:
I originally translated this as well but got rid of it as I wasn't really sure the meager code size benefits were actually worth it (or how much code size it actually saves you in compiled Rust)

aes/aes-soft/src/fixslice32.rs
```diff
@@ -62,12 +62,21 @@ pub(crate) fn aes128_key_schedule(key: &GenericArray<u8, U16>) -> FixsliceKeys12
     }
 
     // Adjust to match fixslicing format
-    for i in (8..72).step_by(32) {
-        inv_shift_rows_1(&mut rkeys[i..(i + 8)]);
-        inv_shift_rows_2(&mut rkeys[(i + 8)..(i + 16)]);
-        inv_shift_rows_3(&mut rkeys[(i + 16)..(i + 24)]);
+    #[cfg(feature = "semi_fixslice")]
```
tarcieri, line 65: It doesn't look like you have a `semi_fixslice` feature declared in Cargo.toml?

I'm also not sure how I feel about a "feature" making things slower.

peterdettman: Sorry, wasn't aware that was a thing. I can add it. As to making things slower, there is the already-existing example of the "no_unroll" feature (in the kuznyechik implementation) that makes the code smaller and slower.

Is it possible in rust to have a public feature like "small_code" that is translated internally in each algorithm to whatever local features make the code smaller?

tarcieri: I wish there were better way to hint at these sort of things in Rust, but unfortunately there aren't AFAIK.

peterdettman:
> I originally translated this as well but got rid of it as I wasn't really sure the meager code size benefits were actually worth it (or how much code size it actually saves you in compiled Rust)

Going by the size of the rlib files in target/release/deps, semi_fixslice reduces code size by about 15%. The performance difference is tiny on my x86_64 machine; an educated guess based on the paper and the code is that it would be ~4% slower on constrained 32-bit devices (even smaller performance loss on arm because of the barrel shifter).

tarcieri:
It might be worth it, but it comes at a cost of more possible codepaths to test and upkeep/maintenance. I think there's also a lot bigger wins to be made in the short term here, like [eliminating the decryption code for AES-CTR use](https://github.com/RustCrypto/traits/pull/352).

Speaking of codepaths to test, can you edit `.github/workflows/aes-soft.yml` and add a line to test the feature in CI?

PR #197, opened 2020-11-12
Rework of xor_columns in fixslice

peterdettman:
- also tweak the lib.rs cfg for fixslice

Results in significant speedup for aes128 and aes256 key schedules.


tarcieri:
Indeed, performance improvements look great!

Issue #10, opened 2017-10-23
Hardware accelerated AES for ARM


newpavlov:
Similar to `aesni` crate it would be nice to have hardware accelerated AES support for ARM using specialized [instructions](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0514g/way1395175472464.html).

quininer:
Should we plan to use stdsimd? https://github.com/rust-lang-nursery/stdsimd/issues/295

newpavlov:
Excellent! I've subscribed to both issues and will try to implement both ARM and x86 versions once they get implemented in the `stdsimd`. (for x86 I [have](https://github.com/RustCrypto/block-ciphers/tree/llvmint_aesni) LLVM intrinsics based draft) One thing which I would like to have is an ability to specify `target_feature` in the `Cargo.toml`, otherwise intrinsics based solution will be less convenient to use, as you'll have to list several features in the `RUSTFLAGS`.

jasondavies:
FYI, I recently added aarch64 AES intrinsics in rust-lang-nursery/stdsimd#398.  I believe AES intrinsics for x86 are in there already.

newpavlov:
Yes, I've seen it, thank you for this addition! When I'll have time I'll try to play with them and implement the crate. (if no one beats me to it that is)

AES-NI intrinsics already utilized in the `aesni` crate, as for Intel  SHA extensions currently I don't have an appropriate CPU to test code on.

tarcieri:
@jack-signal I saw you have hardware accelerated `aarch64` implemented here:

https://github.com/signalapp/libsignal-client/blob/ 946e067/rust/aes-gcm-siv/src/aes/aarch64.rs

Any interest in upstreaming that?

jack-signal:
At this point everything Signal is putting out, including that crate, is AGPL, I doubt there is interest in relicensing so probably wouldn't work. It's worth pointing out though that the linked code is just for AES-256 in forward direction because that is all we needed, and it doesn't implement any of the RustCrypto traits so a ways off from something you could use.

Someone recently pointed me at the MIT licensed https://github.com/shadowsocks/crypto2/blob/master/src/blockcipher/aes/aarch64.rs which might work better as a starting point. But that code doesn't do instruction pipelining, and the key schedule isn't constant time :(

tarcieri:
Ugh, alright, it seems both those approaches are a no-go.

tarcieri:
[...]

PR #200, opened 2020-11-16
Unify the `aes`, `aesni`, and `aes-soft` crates

tarcieri:
Combines all three crates into a single `aes` crate.

The optional `ctr` feature exposes a consistent set of `Aes*Ctr` types as well.

cc @peterdettman


tarcieri:
Note: at present this doesn't fully incorporate all of the tests/benchmarks. I was just trying to get it into a PoC state and get the initial tests passing in CI.

aes/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "aes"
-version = "0.6.0"
-description = "Facade for AES (Rijndael) block ciphers implementations"
+version = "0.7.0-pre"
```
tarcieri, line 3: Note: as a breaking change, I'd like to refactor/ship this concurrently along with the proposed breaking changes to the `cipher` API here:

https://github.com/RustCrypto/traits/pull/352

tarcieri, line 3: I bumped the version worried this might potentially be a breaking change, but I think it may not be?

tarcieri: Changed it back to v0.6.0

tarcieri:
I was a bit torn on whether to call this a breaking change (i.e. bumping the version to `0.7.0-pre`).

It seems like the only breaking case would be treating the types exported from `aes` as the same as either `aesni` or `aes-soft`, which would be breaking if ever AES-NI were/weren't used (depending on what you're expecting).

Given that, I think it's safe to call it backwards compatible. We could potentially release this as v0.6.1, and yank the previous v0.6.0 to prevent any shenanigans.

aes/src/ni.rs
```diff
@@ -71,36 +44,27 @@
...
-    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
-    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg"
-)]
-#![warn(missing_docs, rust_2018_idioms)]
-
-pub use cipher;
-
-#[cfg(feature = "ctr")]
-pub use cipher::stream;
+#![allow(unsafe_code)]
```
newpavlov, line 47: Personally I don't see much sense in using `deny(unsafe_code)` for a whole crate and then enabling it for selected modules.

aes/src/ni/aes192.rs
```diff
@@ -1,11 +1,11 @@
-use crate::arch::*;
+use crate::ni::arch::*;
 use cipher::{
     consts::{U16, U24, U8},
     generic_array::GenericArray,
     BlockCipher, NewBlockCipher,
 };
 
-use crate::utils::{Block128, Block128x8};
+use crate::ni::utils::{Block128, Block128x8};
```
newpavlov, line 8: Alternatively we could use `super` instead of `crate::ni`. It's a bit more compact and will require less changes if we are to move or rename the module.

newpavlov (commented):
I thought that unification is worth only with added runtime detection. Otherwise AES-NI support will be hard to find for users who do not use the `aes` crate directly. In other words most users will use the software implementation. I think I've seen projects which use `aesni` directly to motivate users to enable the `aes` target feature.

But I guess if we'll add the runtime detection before the next minor `cipher` version, merging crate now should not be a problem.

tarcieri:
> I thought that unification is worth only with added runtime detection. 

We definitely need it for that. The current target-based gating of `aesni` vs `aes-soft` crate dependencies makes adding it impossible.

But also it's unclear how to add things like ARM acceleration (#10), which is especially important given the recent launch of the Apple M1. Would that otherwise go into `aesni`? But then we need to gate on two different targets, and things get quite complicated. Add yet another crate? I think as we add more hardware accelerated backends, converging on a single crate makes things a lot easier.

aes/src/ni.rs
tarcieri: Removed the toplevel deny and moved it to the `soft` module in 18a2f0c

aes/src/ni/aes192.rs
tarcieri: Changed all the ones that seem relevant in 2ea345f

aes/src/ni.rs
```diff
@@ -71,36 +44,25 @@
...
-)]
-#![warn(missing_docs, rust_2018_idioms)]
-
-pub use cipher;
-
-#[cfg(feature = "ctr")]
-pub use cipher::stream;
-
-#[cfg(not(feature = "nocheck"))]
-mod target_checks;
```
tarcieri, line 87: It's a bit unclear to me what should happen to this module in a merged crate. Should we remove it?

newpavlov: Yes, the feature and the module will be of no use in the unified crate.

tarcieri: Removed in fa79e60

newpavlov:
>it's unclear how to add things like ARM acceleration

AFAIK an ideal solution is blocked on rust-lang/rfcs#2725. Meanwhile we would have either write platform-specific code to extract features, continue to rely on `RUST_FLAGS`, or add a feature for forcing hardware-accelerated implementation (assuming M1 uses the AArch64 `crypto` extension). Considering that ARM intrinsics are still unstable, I think the latter option will be fine for now.

aes/src/lib.rs
```diff
@@ -1,12 +1,12 @@
...
-//! implementation
-//! - [`aesni`](https://docs.rs/aesni) implementation using
-//! [AES-NI](https://en.wikipedia.org/wiki/AES_instruction_set) instruction set.
-//! Used for x86-64 and x86 target architectures with enabled `aes` and `sse2`
-//! target features (the latter is usually enabled by default).
+//! It provides two different backends based on what target features
+//! are specified:
+//! - "soft" portable constant-time implementation based on [fixslicing].
+//! - [AES-NI] accelerated implementation for `i686`/`x86_64` target
+//!   architectures with enabled `aes` and `sse2` target features
```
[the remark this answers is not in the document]
tarcieri: Removed docs for `sse2` and added some for `ssse3` in f005a97.

aes/src/lib.rs
```diff
@@ -39,28 +39,71 @@
...
+))]
+pub use soft::{Aes128Ctr, Aes192Ctr, Aes256Ctr};
+
+#[cfg(all(
+    feature = "ctr",
+    target_feature = "aes",
+    target_feature = "sse2",
+    any(target_arch = "x86_64", target_arch = "x86"),
+))]
+pub use ni::{Aes128Ctr, Aes192Ctr, Aes256Ctr};
```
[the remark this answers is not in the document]
tarcieri: Added `ssse3` to the gating in 62ef2bb

tarcieri:
Think I addressed all of the outstanding comments.

Re: `cfg-if`-based gating in `lib.rs` I mentioned earlier, I think I'd like to explore that as part of a separate PR.

An additional remaining question is still whether if this is merged, it could be published as v0.6.1, or if we should hold off and do a `cipher` crate upgrade and/or runtime detection first.

newpavlov:
I think we can merge but do not publish just yet.

PR #208, opened 2020-11-26
aes: autodetection support for AES-NI

### LINKED ISSUE #25: AES implementation should be chosen at runtime rather than compile time

tarcieri:
Closes #25.

Adds an off-by-default `autodetect` feature which the `cpuid-bool` crate to detect whether AES-NI is available when compiling on i686/x86_64 architectures.

### DIFF
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -39,7 +39,8 @@ jobs:
       - run: cargo build --release --target ${{ matrix.target }}
       - run: cargo build --release --target ${{ matrix.target }} --features compact
       - run: cargo build --release --target ${{ matrix.target }} --features ctr
-      - run: cargo build --release --target ${{ matrix.target }} --features compact,ctr
+      - run: cargo build --release --target ${{ matrix.target }} --features force-soft
+      - run: cargo build --release --target ${{ matrix.target }} --all-features
 
   # Tests for the portable software backend
   soft:
@@ -73,6 +74,7 @@ jobs:
       - run: cargo test --release --target ${{ matrix.target }}
       - run: cargo test --release --target ${{ matrix.target }} --features compact
       - run: cargo test --release --target ${{ matrix.target }} --features ctr
+      - run: cargo test --release --target ${{ matrix.target }} --features force-soft
       - run: cargo test --release --target ${{ matrix.target }} --all-features
 
   # Tests for the AES-NI backend
@@ -111,6 +113,7 @@ jobs:
       - run: cargo test --release --target ${{ matrix.target }}
       - run: cargo test --release --target ${{ matrix.target }} --features compact
       - run: cargo test --release --target ${{ matrix.target }} --features ctr
+      - run: cargo test --release --target ${{ matrix.target }} --features force-soft
       - run: cargo test --release --target ${{ matrix.target }} --all-features
 
   # Cross-compiled tests
@@ -144,4 +147,5 @@ jobs:
       - run: cross test --release --target ${{ matrix.target }}
       - run: cross test --release --target ${{ matrix.target }} --features compact
       - run: cross test --release --target ${{ matrix.target }} --features ctr
-      - run: cross test --release --target ${{ matrix.target }} --features compact,ctr
+      - run: cross test --release --target ${{ matrix.target }} --features force-soft
+      - run: cross test --release --target ${{ matrix.target }} --all-features
--- a/aes/Cargo.toml
+++ b/aes/Cargo.toml
@@ -23,9 +23,13 @@ opaque-debug = "0.3"
 [dev-dependencies]
 cipher = { version = "=0.3.0-pre", features = ["dev"] }
 
+[target.'cfg(any(target_arch = "x86_64", target_arch = "x86"))'.dependencies]
+cpuid-bool = "0.2"
+
 [features]
-compact = [] # Reduce code size at the cost of performance
+compact = [] # Reduce code size at the cost of slower performance
+force-soft = [] # Disable support for AES hardware intrinsics
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["ctr"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/aes/src/autodetect.rs
+++ b/aes/src/autodetect.rs
@@ -1,0 +1,201 @@
+//! Autodetection support for hardware accelerated AES backends with fallback
+//! to the fixsliced "soft" implementation.
+
+use crate::{Block, ParBlocks};
+use cipher::{
+    consts::{U16, U24, U32, U8},
+    generic_array::GenericArray,
+    BlockCipher, BlockDecrypt, BlockEncrypt, NewBlockCipher,
+};
+
+cpuid_bool::new!(aes_cpuid, "aes");
+
+macro_rules! define_aes_impl {
+    (
+        $name:tt,
+        $module:tt,
+        $key_size:ty,
+        $doc:expr
+    ) => {
+        #[doc=$doc]
+        #[derive(Clone)]
+        pub struct $name {
+            inner: $module::Inner,
+            token: aes_cpuid::InitToken
+        }
+
+        mod $module {
+            #[derive(Copy, Clone)]
+            pub(super) union Inner {
+                pub(super) ni: crate::ni::$name,
+                pub(super) soft: crate::soft::$name,
+            }
+        }
+
+        impl NewBlockCipher for $name {
+            type KeySize = $key_size;
+
+            #[inline]
+            fn new(key: &GenericArray<u8, $key_size>) -> Self {
+                let (token, aesni_present) = aes_cpuid::init_get();
+
+                let inner = if aesni_present {
+                    $module::Inner { ni: crate::ni::$name::new(key) }
+                } else {
+                    $module::Inner { soft: crate::soft::$name::new(key) }
+                };
+
+                Self { inner, token }
+            }
+        }
+
+        impl BlockCipher for $name {
+            type BlockSize = U16;
+            type ParBlocks = U8;
+        }
+
+        impl BlockEncrypt for $name {
+            #[inline]
+            fn encrypt_block(&self, block: &mut Block) {
+                if self.token.get() {
+                    unsafe { self.inner.ni.encrypt_block(block) }
+                } else {
+                    unsafe { self.inner.soft.encrypt_block(block) }
+                }
+            }
+
+            #[inline]
+            fn encrypt_par_blocks(&self, blocks: &mut ParBlocks) {
+                if self.token.get() {
+                    unsafe { self.inner.ni.encrypt_par_blocks(blocks) }
+                } else {
+                    unsafe { self.inner.soft.encrypt_par_blocks(blocks) }
+                }
+            }
+        }
+
+        impl BlockDecrypt for $name {
+            #[inline]
+            fn decrypt_block(&self, block: &mut Block) {
+                if self.token.get() {
+                    unsafe { self.inner.ni.decrypt_block(block) }
+                } else {
+                    unsafe { self.inner.soft.decrypt_block(block) }
+                }
+            }
+
+            #[inline]
+            fn decrypt_par_blocks(&self, blocks: &mut ParBlocks) {
+                if self.token.get() {
+                    unsafe { self.inner.ni.decrypt_par_blocks(blocks) }
+                } else {
+                    unsafe { self.inner.soft.decrypt_par_blocks(blocks) }
+                }
+            }
+        }
+
+        opaque_debug::implement!($name);
+    }
+}
+
+define_aes_impl!(Aes128, aes128, U16, "AES-128 block cipher instance");
+define_aes_impl!(Aes192, aes192, U24, "AES-192 block cipher instance");
+define_aes_impl!(Aes256, aes256, U32, "AES-256 block cipher instance");
+
+#[cfg(feature = "ctr")]
+pub(crate) mod ctr {
+    use super::{aes_cpuid, Aes128, Aes192, Aes256};
+    use cipher::{
+        block::BlockCipher,
+        generic_array::GenericArray,
+        stream::{
+            FromBlockCipher, LoopError, OverflowError, SeekNum, SyncStreamCipher,
+            SyncStreamCipherSeek,
+        },
+    };
+
+    macro_rules! define_aes_ctr_impl {
+        (
+            $name:tt,
+            $cipher:ident,
+            $module:tt,
+            $doc:expr
+        ) => {
+            #[doc=$doc]
+            #[cfg_attr(docsrs, doc(cfg(feature = "ctr")))]
+            pub struct $name {
+                inner: $module::Inner,
+            }
+
+            mod $module {
+                #[allow(clippy::large_enum_variant)]
+                pub(super) enum Inner {
+                    Ni(crate::ni::$name),
+                    Soft(crate::soft::$name),
+                }
+            }
+
+            impl FromBlockCipher for $name {
+                type BlockCipher = $cipher;
+                type NonceSize = <$cipher as BlockCipher>::BlockSize;
+
+                fn from_block_cipher(
+                    cipher: $cipher,
+                    nonce: &GenericArray<u8, Self::NonceSize>,
+                ) -> Self {
+                    let inner = if aes_cpuid::get() {
+                        $module::Inner::Ni(
+                            crate::ni::$name::from_block_cipher(
+                                unsafe { cipher.inner.ni },
+                                nonce
+                            )
+                        )
+                    } else {
+                        $module::Inner::Soft(
+                            crate::soft::$name::from_block_cipher(
+                                unsafe { cipher.inner.soft },
+                                nonce
+                            )
+                        )
+                    };
+
+                    Self { inner }
+                }
+            }
+
+            impl SyncStreamCipher for $name {
+                #[inline]
+                fn try_apply_keystream(&mut self, data: &mut [u8]) -> Result<(), LoopError> {
+                    match &mut self.inner {
+                        $module::Inner::Ni(aes) => aes.try_apply_keystream(data),
+                        $module::Inner::Soft(aes) => aes.try_apply_keystream(data)
+                    }
+                }
+            }
+
+            impl SyncStreamCipherSeek for $name {
+                #[inline]
+                fn try_current_pos<T: SeekNum>(&self) -> Result<T, OverflowError> {
+                    match &self.inner {
+                        $module::Inner::Ni(aes) => aes.try_current_pos(),
+                        $module::Inner::Soft(aes) => aes.try_current_pos()
+                    }
+                }
+
+                #[inline]
+                fn try_seek<T: SeekNum>(&mut self, pos: T) -> Result<(), LoopError> {
+                    match &mut self.inner {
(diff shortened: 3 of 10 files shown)


tarcieri:
Preliminary benchmarks show whatever effect this may have on performance is well within the variance of noise.

Perhaps criterion would give better data but otherwise this appears to have negligible impact on performance.

newpavlov (commented):
I think we can use `union` instead of `enum`. The new version of `cpuid-bool` now supports initialization tokens, which makes the tag in AES types redundant (i.e. the tag will be stored in a static variable inside module created by the new `cpuid-bool` macro).
```rust
cpuid_bool::new(aes_cpuid, "aes");

union Inner {
    ni: crate::ni::Aes128,
    soft: crate::soft::Aes128,
}

pub struct Aes128 {
    inner: Inner,
    token: aes_cpuid::InitToken,
}

fn new(key: &GenericArray<u8, $key_size>) -> Self {
    let (token, val) = aes_cpuid::init_get();
    let inner = if val {
        Inner { ni: crate::ni::Aes128::new(key) }
    } else {
        Inner { soft: crate::soft::Aes128::new(key) }
    };
    Self { token, 
}

fn encrypt_block(&self, block: &mut Block) {
    if self.token.get() {
        unsafe { self.inner.ni.encrypt_block(block) }
    } else {
        unsafe { self.inner.soft.encrypt_block(block) }
    }
}
```

.github/workflows/aes.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - run: ${{ matrix.deps }}
       - run: cargo check --target ${{ matrix.target }} --all-features
       - run: cargo test --release --target ${{ matrix.target }}
+      - run: cargo test --release --target ${{ matrix.target }} --features autodetect
```
newpavlov, line 74: We probably also should specifically test the AES-NI implementation by enabling the `aes` target feature via RUSTFLAGS.

aes/Cargo.toml
```diff
@@ -17,13 +17,15 @@ categories = ["cryptography", "no-std"]
...
 cipher = "=0.3.0-pre"
+cpuid-bool = { version = "0.1", optional = true }
 ctr = { version = "=0.7.0-pre", optional = true }
 opaque-debug = "0.3"
 
 [dev-dependencies]
 cipher = { version = "=0.3.0-pre", features = ["dev"] }
 
 [features]
+autodetect = ["cpuid-bool"] # Detect hardware AES and fallback to "soft" if unavailable
```
newpavlov, line 28: Maybe name it as `x86_autodetect`? Also I wonder if it would be better to enable autodetecton by default and instead use `enforce_soft` feature. In my opinion it will not go against the principle of feature additivity, since I think it's about public API, not inner implementation.

aes/Cargo.toml
```diff
@@ -17,13 +17,15 @@ categories = ["cryptography", "no-std"]
 [dependencies]
 cfg-if = "1"
 cipher = "=0.3.0-pre"
+cpuid-bool = { version = "0.1", optional = true }
```
newpavlov, line 20: Ideally `cpuid-bool` should go to target-specific feature. But I don't remember how exactly optional target-specific features behave...

aes/src/lib.rs
```diff
@@ -68,7 +68,17 @@ cfg_if! {
         any(target_arch = "x86_64", target_arch = "x86"),
     ))] {
         mod ni;
-        pub use ni::{Aes128, Aes192, Aes256};
+
+        cfg_if! {
```
newpavlov, line 72: I think we should join all `cfg_if`s into a single one. The `union`-based proposal will make checking of `target_feature`s here redundant (i.e. with enabled target features the software fallback will be removed from the `autodetect` code making it equivalent to the `ni` implementation).

aes/src/soft/ctr.rs
```diff
@@ -1,5 +1,7 @@
 //! AES in counter mode (a.k.a. AES-CTR)
 
+#![cfg_attr(feature = "autodetect", allow(dead_code))]
+
```
newpavlov, line 4: Is it a temporary thing until you add detection to CTR types?

.github/workflows/aes.yml
tarcieri: It does: https://github.com/RustCrypto/block-ciphers/blob/0fb7bb55fc851db10cade44de9eaf11b08441046/.github/workflows/aes.yml#L82-L85

aes/Cargo.toml
tarcieri: Ideally I'd like to get it working with ARM acceleration too (or at least, have a path to ARM acceleration which doesn't involve renaming features)

aes/src/soft/ctr.rs
tarcieri: Yeah, CTR probably needs similar gating/wrappers. Should've left a TODO.

aes/Cargo.toml
newpavlov: In that case `cpuid-bool` must be a target-specific dependency, otherwise compilation [will fail](https://github.com/RustCrypto/utils/blob/d19e43324aa66e616a46b4a8889f58390eb7be86/cpuid-bool/src/lib.rs#L19-L20) on ARM targets.

newpavlov: FYI: `sha1` and `sha2` use the [`force-soft`](https://github.com/RustCrypto/hashes/blob/master/sha2/Cargo.toml#L39) approach and use `cpuid-bool` as non-optional target-specific dependency. I think it's worth to be consistent and which approach to use here and in other crates.

tarcieri:
@newpavlov I'm not sure that enabling detection without the corresponding target feature being enabled accomplishes anything. See the issue @kazcw opened here:

https://github.com/cryptocorrosion/cryptocorrosion/issues/7

> LLVM has a soft implementation for aesenc for non-AESNI x86-64 CPUs and will gladly use it

If this is true, trying to use AES-NI without the corresponding target feature enabled will use a software fallback provided by LLVM instead.

As a separate issue, I need to test how enabling/disabling AES-NI in UEFI/BIOS affects CPUID.

tarcieri:
> As a separate issue, I need to test how enabling/disabling AES-NI in UEFI/BIOS affects CPUID.

I confirmed disabling AES-NI in UEFI removes the flag from CPUID.

However this is a notable example of how even on the exact same physical CPU, the flag may or may not be present depending on environmental conditions.

tarcieri:
Well, this isn't good. I just ran the test suite against this PR in its current state with the `autodetect` feature enabled on an AES-NI capable host which explicitly has AES-NI disabled in UEFI/BIOS and:

```
  process didn't exit successfully: `/home/tony/block-ciphers/target/debug/deps/aes-eeda3af739ccdef0` (signal: 4, SIGILL: illegal instruction)
```

tarcieri:
@newpavlov as far as I can tell the SIGILL is a bug in `cpuid_bool`, which seems to be returning `true` even when AES-NI is disabled, i.e.:

```
$ lscpu | grep aes
$
```

(I confirmed the flag was present before disabling AES)

I also double checked this is occurring in the autodetection code in this PR, both via a simple `println`-style method:

```
running 1 test
checking CPUID for AES-NI...
YES AES-NI!
```

...and double checking using lldb.

I can try seeing if the problem is solved by using the `is_x86_feature_detected` macro instead...

tarcieri:
I confirmed using the `is_x86_feature_detected` macro fixes the problem and the tests pass.

kazcw:
> @newpavlov I'm not sure that enabling detection without the corresponding target feature being enabled accomplishes anything. See the issue @kazcw opened here:
> 
> [cryptocorrosion/cryptocorrosion#7](https://github.com/cryptocorrosion/cryptocorrosion/issues/7)
> 
> > LLVM has a soft implementation for aesenc for non-AESNI x86-64 CPUs and will gladly use it
> 
> If this is true, trying to use AES-NI without the corresponding target feature enabled will use a software fallback provided by LLVM instead.

@tarcieri I was wrong about the soft fallback in cryptocorrosion/cryptocorrosion#7 -- in the followup comment in that thread I go into what was actually causing the effect I observed.

tarcieri:
@kazcw okay thanks!

I guess it also make sense that if `cpuid-bool` is eagerly omitting the fallback code based on target features, it would always select AES-NI based on their presence.

Will try removing the `target_feature`-based gating then and see if that helps.

tarcieri:
Okay, I confirmed that without supplying any target features, `cpuid-bool` detects AES-NI is disabled and the tests pass.

Let me flip AES-NI back on and determine that it is properly detected when enabled and the performance is what is expected.

tarcieri:
@newpavlov okay, seems everything is working the way you expected.

Let me at least remove the `target_feature`-based gating and push that up.

After that we can then look at inverting the `autodetect` feature into a `force-soft` feature and making `cpuid-bool` gating based on an i686/x86_64 target, and then upgrading `cpuid-bool`.

newpavlov:
Note that with runtime detection we should add `#[target_feature(enable = "..")]` to the code dependent on extensions (and this code must be accessible only after runtime check), otherwise LLVM will not inline intrinsics, which will result in drastically reduced performance.

tarcieri:
@newpavlov this is somewhat problematic:

```
error[E0658]: unions with non-`Copy` fields are unstable
```

I can try marking the inner structs non-`Copy` for now, but I'm a bit wary about that as part of a public API (particularly for `aes::soft`)

newpavlov:
Since `ni` and `soft` types, as well as the `union`s will not be part of the public API it should not be a problem, no? Though we will have to be careful with CTR types, since they have `&mut self` methods.

roblabla:
I'm curious what's the use-case for `force-soft`? It seems odd to me that anyone would want this - if people require using a soft-float aes implementation, they can use the `aes-soft` crate?

newpavlov:
@roblabla 
Currently we plan to deprecate `aes-soft` and `aesni` in favor of the merged `aes` crate.

tarcieri:
Pushed up detection for `Aes*Ctr` as well.

For that I had to use an `enum` (for now at least) as `ctr::Ctr128` doesn't impl `Copy` (or even `Clone`?!)

We can potentially address that, although I'm wary of public `Copy` impls as it makes it way too easy to make copies of keys all over the place, IMO.

tarcieri:
Next steps to me are:

1. Release `cpuid-bool` v0.2
2. Re-enable `cross` tests after above is complete
3. Decide if `enum` is acceptable for `Aes*Ctr` types, and if not, potentially add `Copy` (and at least `Clone`) impls to `ctr::Ctr128`

aes/src/lib.rs
tarcieri: Done

aes/src/soft/ctr.rs
tarcieri: CTR now uses `aes`+`ssse3` runtime detection and this line of code has been removed.

aes/src/autodetect.rs
```diff
@@ -0,0 +1,204 @@
...
+
+            impl FromBlockCipher for $name {
+                type BlockCipher = $cipher;
+                type NonceSize = <$cipher as BlockCipher>::BlockSize;
+
+                fn from_block_cipher(
+                    cipher: $cipher,
+                    nonce: &GenericArray<u8, Self::NonceSize>,
+                ) -> Self {
+                    let (_, aesni_present) = aes_ssse3_cpuid::init_get();
```
newpavlov, line 147: If you don't need the token, you can simply use `aes_ssse3_cpuid::get()`.

newpavlov (commented):
>Release cpuid-bool v0.2

Done.

>Decide if enum is acceptable for Aes*Ctr types, and if not, potentially add Copy (and at least Clone) impls to ctr::Ctr128

Lack of the `Clone` impls is certainly an oversight, but I don't think we should add `Copy`. The main reason is existence of the methods which take `&mut self`, adding `Copy` would make the types too error-prone for my taste.

The best solution for now probably would be to vendor CTR implementation into the `aes` crate until the non-`Copy` unions get stabilized. Luckily the code is not that complex and big.

tarcieri:
> The best solution for now probably would be to vendor CTR implementation into the aes crate until the non-Copy unions get stabilized. 

Wouldn't it be simpler to just stick with an enum for now? It bloats the size of an Aes*Ctr instance marginally, but to me that seems preferable to code duplication.

Note that the `ctr` crate still has an open PR to change the counter behavior, so if we copy and paste it, that becomes three things that need updated (the `ctr` crate, a copy-and-paste version of it in `aes`, and the SSSE3-optimized implementation)

tarcieri:
Alright, this is strange: just tried to re-enable `cross`-based tests and it's now getting linking errors. I confirmed this isn't an environmental issue (#213), so it seems like something legitimately broke the build.

### ARM64

https://github.com/RustCrypto/block-ciphers/pull/208/checks?check_run_id=1480139183

```
  = note: /usr/bin/ld: /home/runner/work/block-ciphers/block-ciphers/target/aarch64-unknown-linux-gnu/release/deps/aes-90b6e8d35dd18b1f.aes.bh0y2qek-cgu.0.rcgu.o: Relocations in generic ELF (EM: 183)
          /home/runner/work/block-ciphers/block-ciphers/target/aarch64-unknown-linux-gnu/release/deps/aes-90b6e8d35dd18b1f.aes.bh0y2qek-cgu.0.rcgu.o: error adding symbols: File in wrong format
          collect2: error: ld returned 1 exit status
```

### PPC32

https://github.com/RustCrypto/block-ciphers/pull/208/checks?check_run_id=1480139204

```
= note: /usr/bin/ld: cannot find Scrt1.o: No such file or directory
          /usr/bin/ld: cannot find crti.o: No such file or directory
          /usr/bin/ld: /home/runner/work/block-ciphers/block-ciphers/target/powerpc-unknown-linux-gnu/release/deps/aes-6dc3112c80dcf8ff.aes.7b1gqfkf-cgu.0.rcgu.o: Relocations in generic ELF (EM: 20)
          /usr/bin/ld: /home/runner/work/block-ciphers/block-ciphers/target/powerpc-unknown-linux-gnu/release/deps/aes-6dc3112c80dcf8ff.aes.7b1gqfkf-cgu.0.rcgu.o: Relocations in generic ELF (EM: 20)
          /usr/bin/ld: /home/runner/work/block-ciphers/block-ciphers/target/powerpc-unknown-linux-gnu/release/deps/aes-6dc3112c80dcf8ff.aes.7b1gqfkf-cgu.0.rcgu.o: Relocations in generic ELF (EM: 20)
          /usr/bin/ld: /home/runner/work/block-ciphers/block-ciphers/target/powerpc-unknown-linux-gnu/release/deps/aes-6dc3112c80dcf8ff.aes.7b1gqfkf-cgu.0.rcgu.o: Relocations in generic ELF (EM: 20)
          /home/runner/work/block-ciphers/block-ciphers/target/powerpc-unknown-linux-gnu/release/deps/aes-6dc3112c80dcf8ff.aes.7b1gqfkf-cgu.0.rcgu.o: error adding symbols: File in wrong format
          collect2: error: ld returned 1 exit status
```

tarcieri:
Okay, managed to get the test suite green again with the `cross` tests running. Strangely they still fail with `--all-features`, though I guess that wasn't tested before.

Will try to get `force-soft` working again.

tarcieri:
Seems `force-soft` was fine and I just screwed up the CI config.

I added it back and `cross`-based tests are now passing.

> Does anyone knows if there are CPUs which have AES-NI, but not SSSE3?

AES-NI was introduced in [Westmere](https://en.wikipedia.org/wiki/Westmere_(microarchitecture)) which also had SSSE3, so I think it's fairly safe to assume any AES-NI capable CPU also has SSSE3.

@newpavlov do you see anywhere that needs to be annotated with `target_feature(enable = "aes")` that isn't already?

I did a quick spot check and it seems like relevant functions are either annotated with that or `#[inline(always)]`, which I assume accomplishes the same thing?

aes/src/autodetect.rs
tarcieri: Updated this to use `aes_cpuid::get()`, as per the Westmere-related comments AFAIK all AES-NI capable CPUs should have SSSE3.

tarcieri:
I'd say this is now ready to merge, so long as the `enum`-based solution for `Aes*Ctr` is at least temporarily acceptable until we can make it work with `union` somehow.

It seems like it might be worth waiting for non-`Copy` union types to stabilize, although I'm not sure what the plan/timeline is for that.

aes/src/autodetect.rs
```diff
@@ -0,0 +1,201 @@
...
+//! to the fixsliced "soft" implementation.
+
+use crate::{Block, ParBlocks};
+use cipher::{
+    consts::{U16, U24, U32, U8},
+    generic_array::GenericArray,
+    BlockCipher, BlockDecrypt, BlockEncrypt, NewBlockCipher,
+};
+
+cpuid_bool::new!(aes_cpuid, "aes");
```
newpavlov, line 11: We either need to add `ssse3`  to this list or use a separate `aes_ssse3_cpuid` module as you did before. Also it may be worth to a comment that SSE2 is implied by AES-NI, so we don't need to check for it separately. Or we could simply add `sse2` to the list to be completely thorough. :)

newpavlov (approved):
Disadvantage of the `enum`-based approach is that we will not be able to remove the fallback branches by enabling the necessary target features via `RUSTFLAGS`. But I guess it can be a good enough starting point and we can revisit it after deciding on what CTR flavors we want to expose in the `ctr` crate.

>do you see anywhere that needs to be annotated with target_feature(enable = "aes") that isn't already?

It looks like you've missed annotations for CTR types and some load/store intrinsics stay outside of the enabled functions. With some shuffling AES-128 parallel encryption improved on my PC from 8 GB/s to 10.6 GB/s. I will try to submit a separate PR with improvements, so I think we can merge this PR as is.

aes/src/autodetect.rs
tarcieri: Per my comments above, as far as I can tell every CPU with AES-NI has *both* SSE2 and SSSE3.

AES-NI was introduced in the [Westmere](https://en.wikipedia.org/wiki/Westmere_(microarchitecture)) architecture, which also has SSSE3. Unless there's some strange AMD/other CPU that has AES-NI but not SSSE3, I don't think it will be a problem.

tarcieri:
@newpavlov I took a look into work on supporting `union` types with non-`Copy` fields in rustc. I think this is the relevant tracking issue?

https://github.com/rust-lang/rust/issues/55149

And this PR appears to stabilize what we'd need, namely:

https://github.com/rust-lang/rust/pull/77547

> It will be possible to have a union with field type ManuallyDrop<T> for any T.

So I think this problem may get addressed soon, after which we can migrate the `enum` to a `union`.

aes/src/autodetect.rs
newpavlov: Per [reference](https://doc.rust-lang.org/reference/attributes/codegen.html#x86-or-x86_64) we can omit `sse2` check if we have verified that `aes` or `ssse3` is enabled, but it does not indicate any implicit dependency between `aes` and `ssse3`. So even though in practice there is probably no such CPU, for our code to be correct we have to check both those features.

tarcieri: Hmm ok, sorry just merged but I can follow up with this.

tarcieri:
I just double checked: `ManuallyDrop` will be stable in 1.49 🎉

PR #27, opened 2018-09-10
Implement IGE (Infinite Garble Extension)

mehcode:
Had to make a breaking change to the BlockModeIv trait as IGE uses a 2-block IV. 

Don't know much about crypto or working in `#[no_std]` crates so my implementation could probably be improved. However, the test vectors pass.

IGE is a block mode used only in Telegram (as far as I know). Described here: https://mgp25.com/AESIGE/

PR #211, opened 2020-11-29
implemented aes-ige block mode

razielgn:
Attempted a rebase of patch #27 by @mehcode.

I've become familiar with the required traits for this patch, but I'm not sure I completely understand all their aspects, especially around those from `generic_array::typenum`.
Let me know how I can improve this work!


tarcieri (approved):
Looks good, thanks!

newpavlov (approved):
I probably will add some minor modifications before publishing, but otherwise looks good!

mehcode:
Thanks for picking this up @razielgn. Happy to see this get merged. I have some old code that was depending on that particular mode (a rust telegram client).

PR #216, opened 2020-12-03
aes: make autodetect unions non-Copy; MSRV 1.49+

tarcieri:
Uses the new `core::mem::ManuallyDrop` feature to allow the inner types of unions to be non-`Copy`.

### DIFF
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.41.0 # MSRV
+          - 1.49.0 # MSRV
           - stable
         target:
           - thumbv7em-none-eabi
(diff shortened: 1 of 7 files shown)


tarcieri:
It works! 🎉 

I wouldn't suggest merging until 1.49 stable is out, but in the meantime, it's nice to see this will be possible soon.

tarcieri:
Rust 1.49 is out 🎉 Merging this.

newpavlov:
It creates a bit of a problem. Previously MSRV was bound by `block-cipher`/`cipher` and was the same for all crates. Also IIUC by enabling the `enofrce-soft` the crate can be compiled on older Rust. Either way we would have to somehow explain this situation in the README. Maybe by adding an MSRV row to the table?

tarcieri:
Yeah we can potentially add and CI for a lower MSRV with the `force-soft` feature enabled.

PR #219, opened 2020-12-31
block-modes: bump MSRV to 1.49+

tarcieri:
The tests are failing because the `aes` crate now uses `ManuallyDrop` unions which are MSRV 1.49 (see #216)


newpavlov:
Hm, strictly speaking MSRV of the crate hasn't changed. Can't we use the `enforce-soft` feature to be able to test the true MSRV?

tarcieri:
@newpavlov good point, yeah that should work

PR #224, opened 2021-02-05
aes: replace AES-NI macros with inline(always) functions

tarcieri:
The previous usage of macros made refactoring extremely difficult, with small changes resulting in hundrdeds upon hundreds of errors when the macros are expanded, as opposed to a single error in a particular function.

This commit replaces the AES-NI `util.rs` macros with `#[inline(always)]` functions.

There is no change to the generated assembly.

### DIFF
--- a/aes/src/ni/utils.rs
+++ b/aes/src/ni/utils.rs
@@ -1,15 +1,17 @@
-#[cfg(test)]
-use super::arch::__m128i;
-#[cfg(test)]
-use core::mem;
+//! Utility functions
 
+use super::arch::*;
 use cipher::{
     consts::{U16, U8},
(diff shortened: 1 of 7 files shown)


aes/src/ni/utils.rs
```diff
@@ -1,15 +1,17 @@
...
+//! Utility functions
 
+use super::arch::*;
 use cipher::{
     consts::{U16, U8},
     generic_array::GenericArray,
 };
 
+#[cfg(test)]
+use core::mem;
```
newpavlov, line 10: You can remove this import and instead use `core::mem::transmute` in the `check` function.

newpavlov (approved):
Looks good!

PR #225, opened 2021-02-07
aes: refactor NI backend

tarcieri:
- Uses array operations on the block buffer
- Uses loops where they don't appear to impact performance

### DIFF
--- a/aes/src/ni/aes128/expand.rs
+++ b/aes/src/ni/aes128/expand.rs
@@ -19,9 +19,9 @@ macro_rules! expand_round {
         t1 = _mm_xor_si128(t1, t3);
         t1 = _mm_xor_si128(t1, t2);
 
-        $enc_keys[$pos] = t1.into();
+        $enc_keys[$pos] = t1;
         let t1 = if $pos != 10 { _mm_aesimc_si128(t1) } else { t1 };
-        $dec_keys[$pos] = t1.into();
+        $dec_keys[$pos] = t1;
     };
 }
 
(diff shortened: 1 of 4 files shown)


tarcieri:
Note: I haven't systematically determined this doesn't change the generated assembly. Preliminary benchmarking shows negligible impact.

aes/src/ni/utils.rs
```diff
@@ -1,5 +1,8 @@
 //! Utility functions
 
+// TODO(tarcieri): check performance impact / generated assembly changes
+#![allow(clippy::needless_range_loop)]
```
tarcieri, line 4: Strangely clippy's suggested refactoring seemed to negatively impact performance?(!)

newpavlov (approved):

aes/src/ni/utils.rs
newpavlov: Yeah, I've seen such behavior as well. Iterator based code probably makes it harder for LLVM to unroll the loops. Also subjectively, I think for such loops range loops result in a better, easier to understand code.

aes/src/ni/aes128/expand.rs
```diff
@@ -19,9 +19,9 @@ macro_rules! expand_round {
         t1 = _mm_xor_si128(t1, t3);
         t1 = _mm_xor_si128(t1, t2);
 
-        _mm_store_si128($enc_keys.as_mut_ptr().offset($pos), t1);
+        $enc_keys[$pos] = t1.into();
         let t1 = if $pos != 10 { _mm_aesimc_si128(t1) } else { t1 };
-        _mm_store_si128($dec_keys.as_mut_ptr().offset($pos), t1);
+        $dec_keys[$pos] = t1.into();
```
newpavlov, line 24: Do you need this `into()`? IIUC `t1` has type `__m128i`, same as the array elements.

tarcieri: Weird, wonder why clippy didn't catch that...

Anyway, fixed in a802938

Issue #226, opened 2021-02-16
Block-modes state saving


jethrogb:
Would you accept a PR adding serde support behind a feature to block-modes? This could be used to implement multi part encryption APIs on top of block-modes. For example: you encrypt a couple of block, save the state, later, restore the state and encrypt a couple more blocks, save/restore, (etc.) finalize. This is a pretty standard paradigm in some legacy crypto APIs that I'd like to implement with a Rust backend.

cc @zugzwang

newpavlov:
Note that a straightforward implementation of the `serde` traits would require `serde` support from block cipher crates as well, which I would like to avoid (e.g. in the case of `aes` it becomes non-trivial, since we support two backends).

I think a better approach would be to add a method for exposing current IV, which then can be used to re-initialize block mode.

jethrogb:
Then you're just pushing the cipher (key) serialization onto the user. You coud just say `C: Serialize + Deserialize`. For block ciphers that don't have it, it's easy to wrap them with something that remembers the key and (de)serializes. Not sure if the bound is necessary for `P`?

Either way is fine for me.

newpavlov:
Not even touching the matter of different backends in `aes`, I think it's far more reasonable to serialize and transfer over wire a key, than a block cipher state, which usually contains a bunch of expanded round keys.

For `ctr` you already can do it by transferring IV/nonce used during mode initialization and current stream position (see [`SyncStreamCipherSeek`](https://docs.rs/cipher/0.2.1/cipher/stream/trait.SyncStreamCipherSeek.html)).

jethrogb:
Well, presumably, a cipher serialization would serialize the original key and not the expanded key. But I suppose it may not be saving the key in order to do that?

tarcieri:
Both backends of the `aes` crate retain only the expanded key, in two different (and incompatible) representations

zugzwang:
@jethrogb it is doable. I can send a PR with the method to expose the "current IV" mentioned by @newpavlov, according to each mode of operation. (The clients can already compute the IV anyway at the cost of XORing blocks - such a method would avoid extra caching).

PR #227, opened 2021-02-18
Add block-modes::IvState trait

zugzwang:
The new IvState trait allows to expose a value that can be used as an IV
for initializing a subsequent BlockMode. These values must be used exclusively
to resume cipher operations, and they MUST NOT be exposed to parties controlling
future blocks.

- Add tests that check correctness of interrupted encryptions (i.e.,
encrypting `k` blocks, initialize a new cipher with the IV returned by
iv_state, and encrypting the remaining blocks).
- Add OFB test.

cc @jethrogb

https://github.com/RustCrypto/block-ciphers/issues/226


zugzwang:
Note: Defined the new trait instead of adding a method to `BlockMode` because some mode states cannot be completely determined by a value comparable to the initial IV (for instance, GOST-CBC, also implementing `BlockMode` needs a positional parameter also).

newpavlov:
Note that you are working on a branch which has migrated to `cipher v0.3`, which is currently under development. So the release of these changes will be blocked on its release. We plan to do it before the end of February. Also we probably will move block mode traits into the `cipher` crate with some re-designs (though I still haven't found a good final design...).

And note that your code is incorrect for the CFB mode. Currently we pre-encrypt IV on initialization, so your `IvState` implementation will simply return an encrypted IV, which when used for re-initialization will result in a doubly encrypted IV. The simplest solution would be to decrypt IV, but a proper one would be to remove the per-encryption, since it's the only mode which does that.

zugzwang:
Thanks @newpavlov for the heads-up on OFB. I wonder now why `ofb_aes128_continued` passes.

jethrogb:
> Also we probably will move block mode traits into the cipher crate with some re-designs (though I still haven't found a good final design...).

What are your thoughts? Is there an issue for this? One thing that's missing today is an update/finish API but it's not too hard to build that on top of encrypt_blocks.

zugzwang:
(Note on last commit: now, CFB doesn't encrypt the IV on initialization and works properly, but needs one IV decryption on parallel CFB decryption. I can refactor CFB parallel decryption to avoid this, if necessary).

newpavlov:
Thank you! As mentioned earlier we will release these changes after the work on `cipher v0.3` is finished.

>What are your thoughts? Is there an issue for this?

No issue I am afraid. I though about exposing the fact that some modes only need block cipher encryption, while others require both encryption and decryption (in `cipher v0.3` we have introduced separate `BlockEncrypt` and `BlockDecrypt` traits). Also I thought about bringing back separate types for encryptors and decryptors. And finally about splitting the trait into two parts: core which works only on blocks (and thus padding-independent) and a buffered wrapper on top of it which would work with slices and will be generic over padding.

jethrogb:
> And finally about splitting the trait into two parts: core which works only on blocks (and thus padding-independent) and a buffered wrapper on top of it which would work with slices and will be generic over padding.

Yes I think this would be helpful

PR #233, opened 2021-04-10
Experimental design

newpavlov:
Partial update based on changes in RustCrypto/traits#589

Later I think it will be worth to deprecate `block-modes` in favor of separate mode crates, we already have some in the `stream-ciphers` repo, so we will only add `cbc`, `ige`, and `pcbc` crates.


newpavlov:
Closing in favor of #284.

PR #257, opened 2021-05-17
aes: expose raw round function API

### LINKED ISSUE #252: aes: expose round function
Exposing the AES round function is of course another level of cryptographic kryptonite beyond the raw block function, but we do have a valid use case.

Deoxys would benefit: https://github.com/RustCrypto/AEADs/pull/311

I think it's possible to build an abstraction for it over the various backends we have.

tarcieri:
Closes #252

Exposes a `hazmat` (more like kryptonite) raw AES round function API intended for use in implementing things like CAESAR candidates (e.g. AEGIS, AEZ, Deoxys)

The current implementation only wraps hardware intrinsics and panics if they aren't available, however the goal is to provide a soft portable fallback as well.

Currently only exposes the "cipher" function (i.e. encryption). However, the goal is to support at least the "equivalent inverse cipher" function (ala Intel AES-NI's `AESDEC`) as well.

cc @zer0x64

### DIFF
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -79,6 +79,7 @@ jobs:
       - run: cargo test --release --target ${{ matrix.target }} --features compact
       - run: cargo test --release --target ${{ matrix.target }} --features ctr
       - run: cargo test --release --target ${{ matrix.target }} --features force-soft
+      - run: cargo test --release --target ${{ matrix.target }} --features hazmat
 
   # Tests for CPU feature autodetection with fallback to portable software implementation
   autodetect:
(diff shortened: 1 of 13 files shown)


tarcieri:
Removing draft/WIP.

I think this is enough to get started, especially for a `hazmat` feature being used to implement more exotic authenticated ciphers.

I propose to merge this as-is and circle back on implementing a software fallback. Unfortunately software fallback is a little bit tricky due to the way the fixsliced implementation optimizes away various operations by coupling them to the key schedule.

However, this should be enough to get started.

PR #261, opened 2021-05-19
kuznyechik: SSE2-based implementation

newpavlov:
Based on https://github.com/aprelev/lg15

This code is several times more performant compared to the existing implementation (135/114 vs. 30/22 MB/s for enc/dec), but uses huge tables (129 KiB in total), which is not great from const time perspective, but unfortunately AFAIK bitsliced Kuznyechik has not been published in open literature yet.

Since the new implementation uses only SSE2 intrinsics, I chose to use only compile-time dispatch.


newpavlov:
Hm, weird. It looks like there has been a compiler bug which causes miscompilation of the SSE2-based implementation. Checking the generated assembly in godbolt, it seems it was fixed only in Rust 1.47. I wonder what should we do in such cases...

I guess the easiest solution for now would be to make the SSE2 implementation feature-gated and disabled by default.

PR #275, opened 2021-07-09
aes: bump ctr dependency to v0.8

newpavlov:


tarcieri (approved):
If you want to cut a breaking release, there are a few other changes I'd like to get in (mostly cleaning up the `hazmat` module which kind of grew organically alongside the `deoxys` crate)

newpavlov:
Nah, I don't plan doing `aes` release in the near future (this is why I've used "unreleased" in the changelog). Also I don't think this change warrants a breaking release. `ctr v0.8` was released only because of the breaking changes in the flavor traits, meanwhile `aes` exposes only already instantiated CTR types. Interface for those is defined via traits from the same `cipher v0.3` (there are also 2 inherent methods, but they are left unchanged from `ctr v0.7`). Thus, I think it's fine to do a patch release here.

PR #284, opened 2021-08-30
Update crates to cipher v0.4

### LINKED ISSUE #210: aes: split encryptors and decryptors
`cipher v0.3` [introduces](https://github.com/RustCrypto/traits/pull/352) separate trait for encryption and decryption. Preliminary migration was done in #202, but AES types have not been split into encryptors and decryptors yet.

### LINKED ISSUE #245: aes: feature flags are not additive
The `aes` crate has two feature flags, both of which _remove_ functionality:

https://github.com/RustCrypto/block-ciphers/blob/f68ad0db49c2136e9d3b3a5d3b2e098782b030a8/aes/Cargo.toml#L29-L31

This is an incorrect use of Cargo feature flags. From my comments at https://github.com/SergioBenitez/cookie-rs/pull/176#issuecomment-834028088, which focus on `force-soft`:

> This seems like an ill-fitted, perhaps incorrect use of features. Features should only be used to _add_ capabilities, but `force-soft` _removes_ the ability to make use of AES-NI. The consequence is that a consumer of `aes` has little control over whether AES-NI is actually used: all it takes is a single dependency, direct or indirect, enabling `force-soft` to disable AES-NI completely, without recourse.
> 
> I would suggest what I would consider to be the "correct" approach: _always_ enable AES-NI if it is known to statically exist, and otherwise, add a feature that adds the ability to dynamically detect its existence, which of course becomes a no-op if the first condition is met.

> Somewhere in the codebase, you already have 2 versions of the code. Let's call them `aes::ni` and `aes::soft`.
> 
> Effectively, we want this:
> 
> ```rust
> fn aes() {
>     #[cfg(have_aesni)]
>     return aes::ni();
> 
>     #[cfg(feature(autodetect)]
>     if detect(aes-ni) { return aes::ni() }
> 
>     aes::soft()
> }
> ```
> 
> This structure makes it clear that `autodetect` is _additive_; it _adds_ the ability to dynamically detect AES-NI and removes nothing otherwise; there is no `not(feature(autodetect))`. This is the only fully correct use of features. If the code that requires `1.49` is restricted to `detect`, then MSRV can remain at 1.41 as long as `autodetect` is not enabled.

The same reasoning applies to the `compact` feature. If it must exist at all, a non-feature `--cfg aes_compact` is the correct approach. This restores control of expected performance to the top-level binary as opposed to any crate in the transitive closure of dependencies.

See also the full comment thread in https://github.com/SergioBenitez/cookie-rs/pull/176.

### LINKED ISSUE #277: [Question/Suggestion] AES encryption examples with text
Hi there.

It seems I am writing a lot of these today, sorry about that.

I am pretty new to using cryptographic functions, unless one exists would it be possible for an example to be added to the online docs to give an example of a round-trip encryption/decryption with text using the various block ciphers?

The one I am looking at is AES though this would be the case with most of them, I'm finding the docs difficult to follow. The example (here: https://docs.rs/aes/0.7.4/aes/) has actually confused me more than it has helped.

This is what I get for looking at a new language and new subject material at the same time.

Thanks.

newpavlov:
Depends on RustCrypto/utils#566 and RustCrypto/traits#727.

Block mode crates are moved into a separate repository, see RustCrypto/block-modes#2.

Closes #210
Closes #245
Closes #277


.github/workflows/aes.yml
```diff
@@ -36,13 +36,25 @@ jobs:
...
+      - env:
+          RUSTFLAGS: "-Dwarnings --cfg aes_compact"
+        run: |
+          cargo build --target ${{ matrix.target }}
+          cargo build --target ${{ matrix.target }} --features hazmat
+      - env:
+          RUSTFLAGS: "-Dwarnings --cfg aes_force_soft --cfg aes_compact"
+        run: |
+          cargo build --target ${{ matrix.target }}
+          cargo build --target ${{ matrix.target }} --features hazmat
```
newpavlov, line 57: I wonder if it's possible to simplify this.

PR #286, opened 2021-09-09
aes: expose helper functions about hardware acceleration detection

vthib:
First of all, thanks for you great work on those crates.

To explain a bit our usecase, we want to be able to use different AEAD algorithms
depending on whether one will use the hardware extension instructions or not.
So, we would like to be able to know if the AES algorithm will be hardware accelerated,
and depending on this information, pick AES-GCM or another algorithm.

Instead of using a cpufeatures check on our side which risks being desynced from
the real check used in the aes crate, exposing a function that returns whether
hw acceleration will be used would be more sound.

Please tell me if this feature is ok with you and/or if you would do it another way,
I tried to put it in the place it made more sense to me, but it's all open to modifications.
Thanks!

### DIFF
--- a/aes/src/autodetect.rs
+++ b/aes/src/autodetect.rs
@@ -17,6 +17,11 @@ use crate::ni as intrinsics;
 
 cpufeatures::new!(aes_intrinsics, "aes");
 
+/// Returns whether hardware acceleration will be used for AES.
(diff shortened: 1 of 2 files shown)


vthib:
I see, thanks for the very detailed answer! I will close this PR then, we'll see how we tackle this another way.

PR #287, opened 2021-09-15
twofish: Derive Clone for Twofish.

teythoon:
This aligns Twofish with AES.  Among other things, it allows its use
with EAX with requires Clone.


tarcieri:
Thanks!

PR #293, opened 2021-12-11
Add implementation of Camellia

sorairolake:
This is implemented based on [RFC 3713](https://datatracker.ietf.org/doc/html/rfc3713) and refer to [Botan](https://github.com/randombit/botan/blob/12fffa08e8167a07756b3564f8c6d25d09458f1c/src/lib/block/camellia/camellia.cpp).
Test vectors are from [NESSIE](https://www.cosic.esat.kuleuven.be/nessie/testvectors/), and tested on all three key sizes of Camellia.

This partially solves #1.


newpavlov:
Thank you for the PR! I will try review it after finishing #284.

sorairolake:
@newpavlov OK.

tarcieri:
@sorairolake sorry we haven't had time to review this. Lots of other things going on and I am not particularly familiar with this cipher

sorairolake:
@tarcieri Should I move this PR to cipher v0.4?
If not, I unmark it a draft.

tarcieri:
Aah. Yeah let's circle back after `cipher` v0.4 is released

newpavlov:
@sorairolake 
Can you please update this PR to cipher v0.4?

sorairolake:
@newpavlov Updated to `cipher` v0.4.
Sorry for the delay.

sorairolake:
@tarcieri Migrated to `cipher` v0.4, so I unmark it a draft.

sorairolake:
@newpavlov Could you review this?

sorairolake:
@tarcieri Migrated to `cipher` v0.4, so I unmarked it a draft.

@newpavlov Could you review this?

camellia/src/camellia.rs
```diff
@@ -0,0 +1,352 @@
...
+
+                BE::write_u64_into(&[d2, d1], block.get_out());
+            }
+        );
+    };
+}
+
+impl_camellia!(Camellia128, 18, U16, "Camellia-128 block cipher instance.");
+impl_camellia!(Camellia192, 24, U24, "Camellia-192 block cipher instance.");
+impl_camellia!(Camellia256, 24, U32, "Camellia-256 block cipher instance.");
```
tarcieri, line 352: It seems like this could be impl'd in terms of a generic core type with (const) generic parameters and type aliases for the specific key sizes, rather than using macros.

Likewise it seems like `gen_subkeys!` could potentially be a function (or rather, two functions) rather than macros.

camellia/src/consts.rs
```diff
@@ -0,0 +1,103 @@
...
+        0xc5, 0x1a, 0xe1, 0xca, 0x47, 0x3d, 0x01, 0xd6, 0x56, 0x4d, 0x0d, 0x66, 0xcc, 0x2d, 0x12,
+        0x20, 0xb1, 0x99, 0x4c, 0xc2, 0x7e, 0x05, 0xb7, 0x31, 0x17, 0xd7, 0x58, 0x61, 0x1b, 0x1c,
+        0x0f, 0x16, 0x18, 0x22, 0x44, 0xb2, 0xb5, 0x91, 0x08, 0xa8, 0xfc, 0x50, 0xd0, 0x7d, 0x89,
+        0x97, 0x5b, 0x95, 0xff, 0xd2, 0xc4, 0x48, 0xf7, 0xdb, 0x03, 0xda, 0x3f, 0x94, 0x5c, 0x02,
+        0x4a, 0x33, 0x67, 0xf3, 0x7f, 0xe2, 0x9b, 0x26, 0x37, 0x3b, 0x96, 0x4b, 0xbe, 0x2e, 0x79,
+        0x8c, 0x6e, 0x8e, 0xf5, 0xb6, 0xfd, 0x59, 0x98, 0x6a, 0x46, 0xba, 0x25, 0x42, 0xa2, 0xfa,
+        0x07, 0x55, 0xee, 0x0a, 0x49, 0x68, 0x38, 0xa4, 0x28, 0x7b, 0xc9, 0xc1, 0xe3, 0xf4, 0xc7,
+        0x9e,
+    ],
+];
```
tarcieri, line 103: Since this is using lookup tables rather than bitslicing, it would probably be good to change the security warning language in README.md to note that it is *NOT* constant time because it's using LUTs.

tarcieri:
Left a few notes. I think this is generally fine but this is not a cipher I'm particularly familiar with so take my review with a grain of salt.

camellia/src/camellia.rs
sorairolake: `gen_subkeys!` has been fixed.

tarcieri:
Going to go ahead and merge this. I think it's acceptable as an initial spike.

I can follow-up on some of my other comments.

sorairolake:
@newpavlov When will this be released?

newpavlov:
Oh, sorry. I haven't done the release for some reason after the merge... I will do it right away.

PR #297, opened 2022-01-22
speck: initial crate

jvdsn:
Based on https://eprint.iacr.org/2013/404.pdf 
Using testvectors from Appendix C of the aforementioned paper.

Notes:
* I implemented this based on the cipher v0.4 branch, so this should only be merged after that branch is merged.
* The crate name is currently `speck`, however, there's already a crate with that name on crates.io, so I think we need to come up with a different name? I've never published on crates.io before.
* The actual code itself uses a macro system similar to AES, with the code inside the macro definition based on SM4.
* I'm not very experienced in Rust, so there might be some things that could be done better. Please let me know!

This partially solves #1.


jvdsn:
With cipher v0.4 being released, this is a bit more ready for review, but of course the crate name issue remains.

newpavlov:
I've registered `speck-cipher`, so we can use it for now.

tarcieri:
I can ask the owners of the current `speck` crate if they'd be willing to give it to us. It's worked before!

jvdsn:
> I can ask the owners of the current `speck` crate if they'd be willing to give it to us. It's worked before!

Perhaps that could be worth a shot, the crate on crates.io doesn't seem to be very active

newpavlov:
>I can ask the owners of the current speck crate if they'd be willing to give it to us.

One problem with `speck` is that it's a post-1.0 crate. So making a breaking `cipher` change would mean that we would have to publish `speck v2.0`.

speck/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+//!
+//! [Speck]: https://en.wikipedia.org/wiki/Speck_(cipher)
+
+#![no_std]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/26acc39f/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/26acc39f/logo.svg",
+    html_root_url = "https://docs.rs/speck/0.0.1"
+)]
+#![deny(unsafe_code)]
```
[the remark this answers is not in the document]
jvdsn: Right, I was following the examples in the other crates, but `forbid` would also work

tarcieri:
@newpavlov if the existing authors are OK with it, we could yank the existing post-1.0 releases and use 0.x releases while our API is unstable, and when we're ready to stabilize, release 2.0.

It doesn't have many downloads (2622) and only one downstream dependency (`speck-cbc`) which itself has no downstream dependencies, so I don't imagine it will be an issue. If anyone complains, we can re-release the original crate under a new name they can switch to.

speck/tests/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+    Speck128_128, Speck128_192, Speck128_256, Speck32_64, Speck48_72, Speck48_96, Speck64_128,
+    Speck64_96, Speck96_144, Speck96_96,
+};
+
+#[test]
+fn speck32_64() {
+    let key = hex!("1918111009080100");
+    let plaintext = hex!("6574694c");
+    let ciphertext = hex!("a86842f2");
+    let cipher = Speck32_64::new(&key.into());
```
newpavlov, line 15: BTW it could be worth to write a macro, so you could define the tests like this:
```rust
new_tests!(
    { Speck32_64, speck32_64, hex!("1918111009080100"), hex!("6574694c"), hex!("a86842f2") }
    { Speck48_72, speck48_72, hex!("1211100a0908020100"), hex!("20796c6c6172"), hex!("c049a5385adc") }
    // ...
)
```

jvdsn:
I added some configuration for a `minimal-versions` action.

speck/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+        #[doc = "block cipher"]
+        #[derive(Clone)]
+        pub struct $name {
+            k: [$word_type; $rounds],
+        }
+
+        impl $name {
+            #[inline]
+            fn from_be_bytes(bytes: [u8; $n / 8]) -> $word_type {
+                // We can't use $word_type::from_be_bytes here because the word size might be different from the word type.
```
newpavlov, line 46: Wouldn't it be better to simply copy bytes into `[u8; size_of::<$word_type>()]` and call `from_be_bytes` on it? I think it also could be a bit more efficient. You also could pass `bytes` as `&[u8]` thus removing the try conversion step (compiler should eliminate all bound checks since all sizes are known at compile time).

Same for the `to_be_bytes` method.

jvdsn: Yes, but we would be making an additional array. Are you sure this would be more efficient?

newpavlov: Yes. Handling of such arrays gets heavily optimized by compiler. See generated assembly: https://rust.godbolt.org/z/MME5f6Wcc

jvdsn: Thank you!

PR #298, opened 2022-01-22
Fix missing no-std category for SM4

jvdsn:
I don't see a reason why `no-std` shouldn't be added here?


newpavlov:
Thank you for noticing this! I have updated categories and keywords in #384, so I will close this PR in it favor.

PR #299, opened 2022-02-04
Update dependencies to build with cargo update -Z minimal-versions

xfix:
It also would be appreciated to release a new version of blockfish with an updated byteorder dependency because it's not a dev-dependency unlike other updated dependencies.

Not committing Cargo.lock is recommended, as libraries cannot really control what dependencies Cargo downloads, and as such it makes sense to test newest versions of dependencies on CI.

### DIFF
--- a/blowfish/Cargo.toml
+++ b/blowfish/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 cipher = "0.3"
-byteorder = { version = "1", default-features = false }
+byteorder = { version = "1.1.0", default-features = false }
 opaque-debug = "0.3"
 
 [dev-dependencies]


tarcieri:
@xfix committing Cargo.lock keeps the build reproducible, and allows us to diagnose how changes in upstream dependencies cause breakages to our project.

This has occurred several times in the past, which is the reason why we check Cargo.lock in now. As an example, we had a user who just so happened to pick up a broken version of `sha2` complain about breakages, and no one could reproduce his problem:

- People who had an older Cargo.lock were fine because they cached an older version
- People doing a clean checkout picked up the latest version

There are pros/cons of either approach and it's a debatable topic. Thus far we have chosen to check in Cargo.lock.

xfix:
Fair enough. Also, from what can I tell `hex-literals 0.3.0` would require higher MSRV, but considering it's a dev-dependency it's not very important, so I will be removing that one.

xfix:
This won't allow `cargo +nightly update -Z minimal-versions` to work within repo itself, but it will work for crates that do depend on `blowfish`, so it's good enough for me.

tarcieri:
Note that we do have a tracking issue for testing with `-Z minimal-versions`: https://github.com/RustCrypto/traits/issues/162

newpavlov:
Thank you!

PR #303, opened 2022-02-16
Fix and test minimal versions build

newpavlov:


.github/workflows/aes.yml
```diff
@@ -56,6 +56,21 @@ jobs:
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        override: true
+        profile: minimal
+    - run: rm ../Cargo.toml
+    - run: rm ../Cargo.lock
+    - run: cargo update -Z minimal-versions
+    - run: cargo test --release
+    - run: cargo test --release --all-features
```
tarcieri, line 72: This seems like something it'd be good to extract as a shared workflow ala https://github.com/RustCrypto/actions/blob/master/.github/workflows/set-msrv.yml

newpavlov: My first attempt in https://github.com/RustCrypto/actions/pull/3 has failed, so I think it's worth to leave it for a later PR.

PR #310, opened 2022-03-16
aes: implement `Zeroize` for aes keys

aticu:
The crate feature `zeroize` was previously present, but unused.

This implements zeroization on drop for all aes types in the software and aesni implementations, if the `zeroize` feature is enabled.

**This PR does not include support for the armv8 implementation.**
I do not have access to hardware to test it and a clean implementation is also likely blocked on the zeroize crate implementing `Zeroize` for the inner `uint8x16_t` type used.

Still I thought this might be a useful patch even without armv8 support.
If you'd prefer, I can also add some documentation describing this limitation to users.

### DIFF
--- a/aes/src/armv8.rs
+++ b/aes/src/armv8.rs
@@ -122,6 +122,9 @@ macro_rules! define_aes_impl {
             }
         }
 
+        #[cfg(feature = "zeroize")]
+        impl zeroize::ZeroizeOnDrop for $name {}
+
         #[doc=$doc]
         #[doc = "block cipher (encrypt-only)"]
         #[derive(Clone)]
@@ -171,6 +174,17 @@ macro_rules! define_aes_impl {
                 f.write_str(stringify!($name_enc))
             }
         }
+
+        impl Drop for $name_enc {
+            #[inline]
+            fn drop(&mut self) {
+                #[cfg(feature = "zeroize")]
+                zeroize::Zeroize::zeroize(&mut self.round_keys);
+            }
+        }
+
+        #[cfg(feature = "zeroize")]
+        impl zeroize::ZeroizeOnDrop for $name_enc {}
 
         #[doc=$doc]
         #[doc = "block cipher (decrypt-only)"]
@@ -234,6 +248,17 @@ macro_rules! define_aes_impl {
                 f.write_str(stringify!($name_dec))
             }
         }
+
+        impl Drop for $name_dec {
+            #[inline]
+            fn drop(&mut self) {
+                #[cfg(feature = "zeroize")]
+                zeroize::Zeroize::zeroize(&mut self.round_keys);
+            }
+        }
+
+        #[cfg(feature = "zeroize")]
+        impl zeroize::ZeroizeOnDrop for $name_dec {}
 
         pub(crate) struct $name_back_enc<'a>(&'a $name_enc);
 
--- a/aes/src/lib.rs
+++ b/aes/src/lib.rs
@@ -159,3 +159,77 @@ use cipher::{
 pub type Block = GenericArray<u8, U16>;
 /// Eight 128-bit AES blocks
 pub type Block8 = GenericArray<Block, U8>;
+
+#[cfg(test)]
+mod tests {
+    #[cfg(feature = "zeroize")]
+    #[test]
+    fn zeroize_works() {
+        use super::soft;
+
+        fn test_for<T: zeroize::ZeroizeOnDrop>(val: T) {
+            use core::mem::{size_of, ManuallyDrop};
+
+            let mut val = ManuallyDrop::new(val);
+            let ptr = &val as *const _ as *const u8;
+            let len = size_of::<ManuallyDrop<T>>();
+
+            unsafe { ManuallyDrop::drop(&mut val) };
+
+            let slice = unsafe { core::slice::from_raw_parts(ptr, len) };
+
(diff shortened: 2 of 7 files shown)


tarcieri:
I could potentially add a feature to `zeroize` to make zeroization on ARMv8 possible.

The necessary intrinsics were stabilized in Rust 1.59.

aticu:
In that case I could add the code required for that, but I wouldn't be able to test it.

tarcieri:
I have an M1 Mac Mini I can test on locally

tarcieri:
Here's a PR to add support to `zeroize`: https://github.com/RustCrypto/utils/pull/749

Turns out I was wrong: only floating point NEON support was stabilized in Rust 1.59. So that feature is nightly-only, but that's fine, because the `armv8` backend for the `aes` crate is also nightly-only.

aes/src/armv8.rs
```diff
@@ -235,6 +249,17 @@ macro_rules! define_aes_impl {
             }
         }
 
+        #[cfg(feature = "zeroize")]
+        impl Drop for $name_dec {
+            fn drop(&mut self) {
+                use cipher::zeroize::Zeroize as _;
+                self.round_keys.zeroize();
+            }
+        }
```
tarcieri, line 258: I think it might be better to define `Drop` unconditionally and gate zeroization in the body:

```suggestion
        impl Drop for $name_dec {
            fn drop(&mut self) {
                #[cfg(feature = "zeroize")]
                zeroize::Zeroize::zeroize(&mut self.round_keys);
            }
        }
```

The same goes for the other impls.

`Drop` is a bit magical so conditionally defining it seems a bit weird. It also changes the bounds.

tarcieri:
I released `zeroize` v1.5.4 with support for zeroizing SIMD registers: https://github.com/RustCrypto/utils/pull/750

Pinning to it will be a bit tricky since it's a transitive dependency of `cipher`, but if you can get things to build with v1.5.4 Cargo.lock we can figure something out (e.g. pinning it in `cipher` and then pinning to a new release of `cipher`)

aticu:
Alright, I've updated the code with the different approach to implementing `Drop`. You're right, I also think this approach is better.

I tried to at least test compiling for armv8, but I wasn't sure how to enable the `aarch64` feature in `zeroize`, since it isn't a direct dependency. Should I maybe add it as a direct dependency instead of using the version in `cipher`? Otherwise the only option I can think of is to add a feature for that in `cipher` that can then be enabled in `aes`.
Do you know of a way to enable features in transitive dependencies?

tarcieri:
Aah yeah, you'll probably want to use a direct dependency

tarcieri:
Hmm, weird, the CI workflow isn't running properly due to a syntax error: https://github.com/RustCrypto/block-ciphers/actions/runs/1998929396/workflow

tarcieri:
Already fixed it: #311

Can you rebase?

aticu:
Done

aes/src/lib.rs
```diff
@@ -159,3 +159,77 @@ use cipher::{
...
+        test_for(soft::Aes128Enc::new(&key_128));
+        test_for(soft::Aes128Dec::new(&key_128));
+        test_for(soft::Aes192::new(&key_192));
+        test_for(soft::Aes192Enc::new(&key_192));
+        test_for(soft::Aes192Dec::new(&key_192));
+        test_for(soft::Aes256::new(&key_256));
+        test_for(soft::Aes256Enc::new(&key_256));
+        test_for(soft::Aes256Dec::new(&key_256));
+
+        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
tarcieri, line 199:
```suggestion
        #[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), not(feature = "force-soft")))]
```

tarcieri (approved):

aes/src/armv8.rs
```diff
@@ -172,6 +175,16 @@ macro_rules! define_aes_impl {
             }
         }
 
+        impl Drop for $name_enc {
+            fn drop(&mut self) {
```
newpavlov, line 179: It could be worth to add `#[inline]` attribute here and in other places. Otherwise, without LTO Rust will not be able to inline it.

newpavlov (approved):

aes/src/armv8.rs
aticu: As far as I'm aware the only difference between implementing `Drop` and not implementing it is being able to implement `Copy`. The [reference](https://doc.rust-lang.org/stable/reference/special-types-and-traits.html?highlight=drop#drop) also does not seem to say much more about that.

The only other differences I could think of are performance issues, such as not having to track drop flags and not running an empty destructor (though inlining should prevent that). I added the inline attribute to all `drop` functions now.

In general I agree with @tarcieri that conditionally implementing `Drop` (or any trait) based on a not obviously related feature could lead to problems.

tarcieri: If nothing else the bounds are different, although whether `Drop` bounds should be used/allowed at all is a matter of debate. Nevertheless, they do.

The other major difference is it's not possible to move out of a struct which impls `Drop`, although that's not happening in the current implementation so it's not a major issue.

I think `Drop` is magical enough conditionally dropping based on whether or not features are impl'd is a bit weird.

aticu:
Thank you for the responses and help in getting this merged so quickly!

PR #311, opened 2022-03-17
Fix `aes` and `kuznyechik` CI workflows

tarcieri:
Looks like they were broken in #284


.github/workflows/aes.yml
```diff
@@ -3,6 +3,7 @@ name: aes
 on:
   pull_request:
     paths:
+      - ".github/workflows/aes.yml"
```
tarcieri, line 6: We should probably do this everywhere in order to ensure changes to a workflow trigger the workflow

kuznyechik/src/soft/mod.rs
```diff
@@ -81,7 +81,7 @@ impl AlgorithmName for Kuznyechik {
 #[cfg_attr(docsrs, doc(cfg(feature = "zeroize")))]
 impl Drop for Kuznyechik {
     fn drop(&mut self) {
-        self.keys.zeroize();
+        self.keys.iter_mut().for_each(|key| key.zeroize());
```
tarcieri, line 84: Unfortunately the `Zeroize` impl on `core::slice::IterMut` couldn't handle this case, I believe because this explicit version works due to implicit deref coercion.

PR #317, opened 2022-04-29
ParBlocksSizeUser impls

rozbb:
This impls `ParBlocksSizeUser` for AES and Kuznyechik block ciphers. `ParBlocksSizeUser` was already impl'd for their backends, so this was a matter of copy-pasting.

This unblocks the fix for https://github.com/RustCrypto/AEADs/issues/410

### DIFF
--- a/aes/src/armv8.rs
+++ b/aes/src/armv8.rs
@@ -110,6 +110,10 @@ macro_rules! define_aes_impl {
             }
         }
 
+        impl ParBlocksSizeUser for $name {
+            type ParBlocksSize = U8;
+        }
+
         impl fmt::Debug for $name {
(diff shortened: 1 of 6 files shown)


newpavlov:
No, it's meaningless to implement `ParBlocksSizeUser` on ciphers themselves. If you need access to explicit parallel block processing, then you SHOULD use the `with_backend` methods.

tarcieri:
@newpavlov can you take a look at https://github.com/RustCrypto/AEADs/issues/410 ?

None of us can figure out how to properly do the upgrade.

newpavlov:
@tarcieri Yes, I will try to do it this weekend.

tarcieri:
Seems this is not necessary

PR #322, opened 2022-08-16
Gift

Schmid7k:
This pull request contains a constant-time software fixslicing implementation of the [GIFT](https://eprint.iacr.org/2017/622.pdf) block cipher, based on the C implementation of the original authors found [here](https://github.com/aadomn/gift/tree/master/crypto_bc/gift128/opt32). 

Gift is a PRESENT based block cipher with focus on energy efficiency and a small memory footprint making it ideal for usage in resource constrained environments. It gained some popularity as part of GIFT-COFB, which is a finalist in the current [NIST lightweight cryptography competition](https://csrc.nist.gov/Projects/lightweight-cryptography) as well as SUNDAE-GIFT.

This pull request comes with an implementation of GIFT-128, meaning Gift operating on 128-bit blocks, though I am planning on implementing GIFT-64 in the future.

It passes all test vectors and performs at ~44.5 cpb on an Intel Core i7 8700k with 3.7GHz core clock, while the C version "only" performs at ~47.1 cpb on the same machine.


Schmid7k:
And last but not least, Gift also rebased from upstream.

tarcieri (approved):

PR #340, opened 2022-10-05
aria: initial implementaion of the ARIA cipher (RFC 5794)

lumag:
Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>


aria/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+        #[cfg(feature = "zeroize")]
+        #[cfg_attr(docsrs, doc(cfg(feature = "zeroize")))]
+        impl ZeroizeOnDrop for $name {}
+
+        cipher::impl_simple_block_encdec!(
+            $name, U16, cipher, block,
+            encrypt: {
+                let b = block.get_in();
+
+                let mut p0 = u128::from_be_bytes(b[0..16].try_into().unwrap());
```
newpavlov, line 372: `b` has type `&GenericArray<u8, U16>`, so you should be able to use `Into` for converting it into `[u8; 16].`

aria/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+                    p1 = fo(p0 ^ cipher.ek[i]);
+                    p0 = fe(p1 ^ cipher.ek[i + 1]);
+                }
+
+                let p1 = fo(p0 ^ cipher.ek[$subkey_size - 3]);
+                let c = sl2(p1 ^ cipher.ek[$subkey_size - 2]) ^ cipher.ek[$subkey_size - 1];
+
+                let block = block.get_out();
+
+                block.copy_from_slice(&c.to_be_bytes());
```
newpavlov, line 385: Join these two lines?
```rust
block.get_out().copy_from_slice(&c.to_be_bytes());
```

aria/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+use crate::consts::{C1, C2, C3, SB1, SB2, SB3, SB4};
+
+macro_rules! diffuse {
+    ($x0: expr, $x1: expr, $x2: expr, $x3: expr,
+     $x4: expr, $x5: expr, $x6: expr, $x7: expr,
+     $x8: expr, $x9: expr, $x10: expr, $x11: expr,
+     $x12: expr, $x13: expr, $x14: expr, $x15: expr,) => {
+        let x = [
+            0x0000_0001_0100_0100_0101_0000_0001_0100 * $x0 as u128,
+            0x0000_0100_0001_0001_0101_0000_0100_0001 * $x1 as u128,
```
newpavlov, line 64: You can replace this whole macro with the following function:
```rust
#[inline(always)]

pub fn diffuse(x: [u8; 16]) -> u128 {
    DIFFUSE_CONSTS.iter().zip(x).map(|(a, b)| a * b as u128).fold(0, |a, v| a ^ v)
}
```
Compiler is [smart enough](https://rust.godbolt.org/z/a5ceG6heq) to unroll everything.

newpavlov (commented):
It could be worth to split encryptors and decryptors into separate types, but it can be done in a later PR.

aria/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+        .map(|(a, b)| a * b as u128)
+        .fold(0, |a, v| a ^ v)
+}
+
+fn a(x128: u128) -> u128 {
+    let x = x128.to_be_bytes();
+    diffuse([
+        x[0], x[1], x[2], x[3], x[4], x[5], x[6], x[7], x[8], x[9], x[10], x[11], x[12], x[13],
+        x[14], x[15],
+    ])
```
newpavlov, line 71: You can simply write it as `diffuse(x.to_be_bytes())`.

lumag: done

lumag:
@newpavlov gracious ping

newpavlov (approved):
Sorry for the late review! I will release it right away.

PR #346, opened 2022-11-23
initial rc5 impl

antonio-dropulic:
Implemented RC5 based on the revised paper.  Got test vector at: https://www.ietf.org/archive/id/draft-krovetz-rc6-rc5-vectors-00.txt

Core impl features a generic impl using typenum and generic array. Const generic implementation
without generic expressions doesn't seem appropriate.

I was not sure what api would work best with the cipher crate so i decided to propagate the InOut ptr.
Put all the core fn behind a trait to avoid rewriting generic bounds.

For now only 32, 20, 20 parametrization made public.


rc5/src/block_cipher.rs
```diff
@@ -0,0 +1,64 @@
+use cipher::consts::{U16, U20, U8};
+
+use crate::core::{ExpandedKeyTable, RC5};
+use cipher::{impl_simple_block_encdec, AlgorithmName, KeyInit};
+use cipher::{inout::InOut, Block, BlockCipher, KeySizeUser};
+
+pub struct RC5_32_20_16 {
```
lumag, line 7: The choice of 20 rounds is not obvious. Both OpenSSL and RFC 2040 default to 12 rounds.

rc5/src/block_cipher.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn new(key: &cipher::Key<Self>) -> Self {
+        Self {
+            key_table: Self::substitute_key(key),
+        }
+    }
+}
+
+impl AlgorithmName for RC5_32_20_16 {
+    fn write_alg_name(f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str("RC5-32/12/16")
```
lumag, line 39: So, 20 or 12?

rc5/src/block_cipher.rs
antonio-dropulic: from the: https://www.ietf.org/archive/id/draft-krovetz-rc6-rc5-vectors-00.txt

" The original RC5 publication suggested using 12 rounds when w=32 and
   16 rounds when w=64.  In response to cryptanalysis, the authors
   changed the recommendation when w=32 to 16 rounds [RC5sec].  NIST
   concluded in 2001 that RC6 with w=32 and r=20 "appears to have an
   adequate security margin" [NIST] which agrees with the RC6 inventor's
   analysis [RC6sec].  No other recommendations have been published
   regarding the number of rounds needed for good security with other w
   values."
   
   I'm not sure what should be the defaults.

tarcieri: Exposing them as generic parameters makes sense to me

antonio-dropulic: I've added the generic implementation, but the compiler seems to strugle with optimizing it. I notice a 3x penalty in performance compared to the previous version.

```
 name                         control ns/iter    generic ns/iter    diff ns/iter   diff %  speedup 
 rc5_32_12_16_decrypt_block   14,611 (560 MB/s)  63,711 (128 MB/s)        49,100  336.05%   x 0.23 
 rc5_32_12_16_decrypt_blocks  14,659 (558 MB/s)  58,842 (139 MB/s)        44,183  301.41%   x 0.25 
 rc5_32_12_16_encrypt_block   11,242 (728 MB/s)  51,916 (157 MB/s)        40,674  361.80%   x 0.22 
 rc5_32_12_16_encrypt_blocks  11,467 (714 MB/s)  51,360 (159 MB/s)        39,893  347.89%   x 0.22 
 rc5_32_16_16_decrypt_block   19,621 (417 MB/s)  77,881 (105 MB/s)        58,260  296.93%   x 0.25 
 rc5_32_16_16_decrypt_blocks  19,800 (413 MB/s)  76,284 (107 MB/s)        56,484  285.27%   x 0.26 
 rc5_32_16_16_encrypt_block   17,708 (462 MB/s)  81,863 (100 MB/s)        64,155  362.29%   x 0.22 
 rc5_32_16_16_encrypt_blocks  17,704 (462 MB/s)  75,664 (108 MB/s)        57,960  327.38%   x 0.23 
 ```
 
Any idea why this might be?

antonio-dropulic: Upon closer inspection i found that the generic version would not inline calls to word functions, while the non generic one would. After adding inline annotations to Word impl the performance is once again comparable.

axos88:
Can we get this merged and released? Is it usable @antonio-dropulic? I'd very much like to use it, and would be happy to do from your branch until it gets merged.

tarcieri (approved):

Issue #351, opened 2023-02-19
`encrypt_block` vs `encrypt_blocks`, what is 'parallel block processing'?


KaneGreen:
I am recently learning how to use `aes`. I'm reading the [docs](https://docs.rs/aes/0.8.2/aes/).  
Now I note `encrypt_block` can process a single block, while `encrypt_blocks` can process many blocks.  
In the docs' example, there is a comment:
> implementation supports parallel block processing
> number of blocks processed in parallel depends in general on hardware capabilities

I'm wondering, is this *parallel* an automatic multi-thread implement? And the `hardware capabilities` is the CPU core number?

tarcieri:
It leverages [instruction-level parallelism](https://en.wikipedia.org/wiki/Instruction-level_parallelism), which allows a single CPU to compute multiple AES blocks in parallel.

It does not require multiple CPU cores.

KaneGreen:
Thank you.

However, I have an advice: maybe it will be better if some sentences like "this is not by multi-threading" can be added in the docs' comment mentioned above.  
Because I think when users see "parallel", they are more likely to first think of multi-threaded parallelism.

-----------
Besides, I have two more question:
* I have run `cargo bench` on several X86 computers. It seems when AES-NI is disabled, there was always nearly 4x speed-up for `encrypt_blocks`, but just a little speed-up when AES-NI is available. So, does this "parallelism" mainly come from bit-slicing?
* Will I still benefit from the nearly 4x speed-up when I use high-level APIs such as `cbc::Encryptor<aes::Aes128>` and `aes_gcm::Aes256Gcm` if AES-NI is disabled?

tarcieri:
The parallelism comes from the CPU working on multiple blocks in parallel. You should see a roughly 8X speedup versus a single block.

It should be roughly an order of magnitude faster than the bitsliced implementation. That said, the bitsliced implementation computes 4 blocks in parallel on 64-bit architectures.

CBC encryption will not see a parallelization speedup since each block has a data dependency on the previous block. CBC decryption, however, is parallelizable.

GCM is fully parallelizable for both encryption and decryption.

If you’d like to submit a docs PR with your suggestions, that’d be great.

KaneGreen:
Docs PR is ready.

-----
But I can't see the "roughly 8X speedup" you said.

AES-NI enabled: 5%~10% speedup
`RUSTFLAGS="-C target-feature=+aes,+ssse3"`
```
[...]

PR #352, opened 2023-02-20
`aes` docs update

KaneGreen:
* Added a note about instruction-level parallelism to Examples. (#351)
* Added a space to `-Ctarget-feature=+aes,+ssse3` since having a space after `-C` seems to be the [more common form](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-feature).
(`-Dwarnings` are not changed, because is seems [ignoring space after `-D` is more common](https://doc.rust-lang.org/clippy/usage.html#command-line))


newpavlov:
Thank you!

PR #356, opened 2023-03-11
Replace unmaintained actions-rs/cargo action in CI workflow

striezel:
Basically all of the `actions-rs/*` actions are unmaintained. See <https://github.com/actions-rs/toolchain/issues/216> for more information. Due to their age they generate several warnings in CI runs, for example in https://github.com/RustCrypto/block-ciphers/actions/runs/4394018647:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/cargo@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

To get rid of those warnings the occurrences of `actions-rs/cargo` are replaced by direct invocations of `cargo`.

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -33,7 +33,4 @@ jobs:
           toolchain: stable
           components: rustfmt
       - name: Run cargo fmt
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+        run: cargo fmt --all -- --check


tarcieri:
So this was missed in #354 I take it?

striezel:
> So this was missed in #354 I take it?

Not really. #354 takes care of `actions-rs/toolchain` while this PR handles `actions-rs/cargo`. That's a different action, it just happens to be by the same person / organization and suffers the same problems. I usually create two PRs for the two actions, because most projects I encountered use both actions quite extensively, so it's easier to do that in two steps.

However, in this particular case, there was just a single occurrence of `actions-rs/cargo` so I could have replaced that together with `actions-rs/toolchain`. If you prefer it that way, I can replace both in a single PR in the future, too.

tarcieri:
Yeah it'd probably be good to just rip out everything `actions-rs` related in one go, especially as we have dozens of repos

striezel:
Alright, I'll replace `actions-rs/toolchain` and `actions-rs/cargo` in one go for the remaining RustCrypto repositories.

PR #365, opened 2023-06-15
aes: support `aes_armv8` on Rust 1.61+ using `asm!`

tarcieri:
Adds "polyfills" for the unstable ARMv8 AES intrinsics using the `asm!` macro which was stabilized in Rust 1.59. However note we also need `target_feature` stabilizations for `aes` and `neon` which occurred in Rust 1.61.

Based on benchmarks this has no effect on performance, although it was necessary to place AESE/AESMC and AESD/AESIMC into a single `asm!` block in order to ensure that instructions fuse properly, as they did when using the proper intrinsics.

In the next breaking release, we should be able to get rid of the `aes_armv8` configuration parameter entirely, bumping MSRV to 1.59 and then ARMv8 support should Just Work(TM) where available.

Performance appears to be unchanged.

### Before

```
$ RUSTFLAGS="--cfg aes_armv8" cargo +nightly bench
   Compiling aes v0.8.2 (/Users/tony/src/RustCrypto/block-ciphers/aes)
    Finished bench [optimized] target(s) in 2.24s
     Running unittests src/lib.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/aes-387e2e27bf9fddbd)

running 4 tests
test armv8::test_expand::aes128_key_expansion ... ignored
test armv8::test_expand::aes128_key_expansion_inv ... ignored
test armv8::test_expand::aes192_key_expansion ... ignored
test armv8::test_expand::aes256_key_expansion ... ignored

test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/mod.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/mod-8400f88c92c19848)

running 15 tests
test aes128_decrypt_block  ... bench:       1,059 ns/iter (+/- 8) = 15471 MB/s
test aes128_decrypt_blocks ... bench:       1,189 ns/iter (+/- 17) = 13779 MB/s
test aes128_encrypt_block  ... bench:       1,056 ns/iter (+/- 31) = 15515 MB/s
test aes128_encrypt_blocks ... bench:       1,192 ns/iter (+/- 86) = 13744 MB/s
test aes128_new            ... bench:         102 ns/iter (+/- 0)
test aes192_decrypt_block  ... bench:       1,510 ns/iter (+/- 10) = 10850 MB/s
test aes192_decrypt_blocks ... bench:       1,272 ns/iter (+/- 84) = 12880 MB/s
test aes192_encrypt_block  ... bench:       1,448 ns/iter (+/- 22) = 11314 MB/s
test aes192_encrypt_blocks ... bench:       1,276 ns/iter (+/- 24) = 12840 MB/s
test aes192_new            ... bench:         103 ns/iter (+/- 1)
test aes256_decrypt_block  ... bench:       1,728 ns/iter (+/- 31) = 9481 MB/s
test aes256_decrypt_blocks ... bench:       1,545 ns/iter (+/- 10) = 10604 MB/s
test aes256_encrypt_block  ... bench:       1,727 ns/iter (+/- 14) = 9486 MB/s
test aes256_encrypt_blocks ... bench:       1,547 ns/iter (+/- 23) = 10590 MB/s
test aes256_new            ... bench:         124 ns/iter (+/- 3)

test result: ok. 0 passed; 0 failed; 0 ignored; 15 measured; 0 filtered out; finished in 24.93s
```

### After

```
$ RUSTFLAGS="--cfg aes_armv8" cargo +nightly bench
   Compiling aes v0.8.2 (/Users/tony/src/RustCrypto/block-ciphers/aes)
    Finished bench [optimized] target(s) in 1.07s
     Running unittests src/lib.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/aes-eb55ebb5f7ca4ca0)

running 4 tests
test armv8::test_expand::aes128_key_expansion ... ignored
test armv8::test_expand::aes128_key_expansion_inv ... ignored
test armv8::test_expand::aes192_key_expansion ... ignored
test armv8::test_expand::aes256_key_expansion ... ignored

test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/mod.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/mod-6305260e7c7483e7)

running 15 tests
test aes128_decrypt_block  ... bench:       1,060 ns/iter (+/- 7) = 15456 MB/s
test aes128_decrypt_blocks ... bench:       1,025 ns/iter (+/- 13) = 15984 MB/s
test aes128_encrypt_block  ... bench:       1,059 ns/iter (+/- 7) = 15471 MB/s
test aes128_encrypt_blocks ... bench:       1,024 ns/iter (+/- 8) = 16000 MB/s
test aes128_new            ... bench:         101 ns/iter (+/- 4)
test aes192_decrypt_block  ... bench:       1,446 ns/iter (+/- 16) = 11330 MB/s
test aes192_decrypt_blocks ... bench:       1,271 ns/iter (+/- 25) = 12890 MB/s
test aes192_encrypt_block  ... bench:       1,447 ns/iter (+/- 18) = 11322 MB/s
test aes192_encrypt_blocks ... bench:       1,275 ns/iter (+/- 8) = 12850 MB/s
test aes192_new            ... bench:         103 ns/iter (+/- 5)
test aes256_decrypt_block  ... bench:       1,833 ns/iter (+/- 17) = 8938 MB/s
test aes256_decrypt_blocks ... bench:       1,569 ns/iter (+/- 9) = 10442 MB/s
test aes256_encrypt_block  ... bench:       1,727 ns/iter (+/- 19) = 9486 MB/s
test aes256_encrypt_blocks ... bench:       1,570 ns/iter (+/- 18) = 10435 MB/s
test aes256_new            ... bench:         124 ns/iter (+/- 1)

test result: ok. 0 passed; 0 failed; 0 ignored; 15 measured; 0 filtered out; finished in 21.59s
```

cc @codahale

### DIFF
--- a/aes/src/armv8/expand.rs
+++ b/aes/src/armv8/expand.rs
@@ -2,6 +2,10 @@
 
 use core::{arch::aarch64::*, mem, slice};
 
+// Stable "polyfills" for unstable core::arch::aarch64 intrinsics
+// TODO(tarcieri): remove when these intrinsics have been stabilized
+use super::intrinsics::{vaeseq_u8, vaesimcq_u8};
+
 /// There are 4 AES words in a block.
 const BLOCK_WORDS: usize = 4;
 
--- a/aes/src/armv8/expand.rs
+++ b/aes/src/armv8/expand.rs
@@ -57,7 +57,6 @@ pub(super) fn expand_key<const L: usize, const N: usize>(key: &[u8; L]) -> [uint
 /// This is the reverse of the encryption keys, with the Inverse Mix Columns
 /// operation applied to all but the first and last expanded key.
 #[target_feature(enable = "aes")]
-#[target_feature(enable = "neon")]
 pub(super) unsafe fn inv_expanded_keys<const N: usize>(expanded_keys: &mut [uint8x16_t; N]) {
     assert!(N == 11 || N == 13 || N == 15);
 
@@ -70,7 +69,6 @@ pub(super) unsafe fn inv_expanded_keys<const N: usize>(expanded_keys: &mut [uint
 
 /// Sub bytes for a single AES word: used for key expansion.
 #[target_feature(enable = "aes")]
-#[target_feature(enable = "neon")]
 unsafe fn sub_word(input: u32) -> u32 {
     let input = vreinterpretq_u8_u32(vdupq_n_u32(input));
 
--- a/aes/src/armv8/expand.rs
+++ b/aes/src/armv8/expand.rs
@@ -41,9 +41,9 @@ pub(super) fn expand_key<const L: usize, const N: usize>(key: &[u8; L]) -> [uint
         let mut word = ek_words[i - 1];
 
         if i % nk == 0 {
-            word = sub_word(word).rotate_right(8) ^ ROUND_CONSTS[i / nk - 1];
+            word = unsafe { sub_word(word) }.rotate_right(8) ^ ROUND_CONSTS[i / nk - 1];
         } else if nk > 6 && i % nk == 4 {
-            word = sub_word(word)
+            word = unsafe { sub_word(word) };
         }
 
         ek_words[i] = ek_words[i - nk] ^ word;
@@ -56,26 +56,26 @@ pub(super) fn expand_key<const L: usize, const N: usize>(key: &[u8; L]) -> [uint
 ///
 /// This is the reverse of the encryption keys, with the Inverse Mix Columns
 /// operation applied to all but the first and last expanded key.
-#[inline]
-pub(super) fn inv_expanded_keys<const N: usize>(expanded_keys: &mut [uint8x16_t; N]) {
+#[target_feature(enable = "aes")]
+#[target_feature(enable = "neon")]
+pub(super) unsafe fn inv_expanded_keys<const N: usize>(expanded_keys: &mut [uint8x16_t; N]) {
     assert!(N == 11 || N == 13 || N == 15);
 
     for ek in expanded_keys.iter_mut().take(N - 1).skip(1) {
-        unsafe { *ek = vaesimcq_u8(*ek) }
+        *ek = vaesimcq_u8(*ek);
     }
 
     expanded_keys.reverse();
 }
 
 /// Sub bytes for a single AES word: used for key expansion.
-#[inline(always)]
-fn sub_word(input: u32) -> u32 {
-    unsafe {
-        let input = vreinterpretq_u8_u32(vdupq_n_u32(input));
+#[target_feature(enable = "aes")]
+#[target_feature(enable = "neon")]
+unsafe fn sub_word(input: u32) -> u32 {
+    let input = vreinterpretq_u8_u32(vdupq_n_u32(input));
 
-        // AES single round encryption (with a "round" key of all zeros)
-        let sub_input = vaeseq_u8(input, vdupq_n_u8(0));
+    // AES single round encryption (with a "round" key of all zeros)
+    let sub_input = vaeseq_u8(input, vdupq_n_u8(0));
 
-        vgetq_lane_u32(vreinterpretq_u32_u8(sub_input), 0)
-    }
+    vgetq_lane_u32(vreinterpretq_u32_u8(sub_input), 0)
 }
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -219,7 +219,7 @@ jobs:
           cross test --package aes --target ${{ matrix.target }}
           cross test --package aes --target ${{ matrix.target }} --features hazmat
 
-  # ARMv8 cross-compiled tests for AES intrinsics (nightly-only)
+  # ARMv8 cross-compiled tests for AES intrinsics
   armv8:
     env:
       RUSTFLAGS: "-Dwarnings --cfg aes_armv8"
@@ -227,7 +227,7 @@ jobs:
       matrix:
         include:
           - target: aarch64-unknown-linux-gnu
-            rust: nightly
+            rust: 1.61.0 # MSRV for `aes_armv8`
     runs-on: ubuntu-latest
     # Cross mounts only current package, i.e. by default it ignores workspace's Cargo.toml
     defaults:
--- a/aes/src/armv8.rs
+++ b/aes/src/armv8.rs
@@ -223,7 +223,7 @@ macro_rules! define_aes_impl {
         impl From<&$name_enc> for $name_dec {
             fn from(enc: &$name_enc) -> $name_dec {
                 let mut round_keys = enc.round_keys;
-                inv_expanded_keys(&mut round_keys);
+                unsafe { inv_expanded_keys(&mut round_keys) };
                 Self { round_keys }
             }
         }
--- a/aes/src/armv8.rs
+++ b/aes/src/armv8.rs
@@ -14,6 +14,7 @@ pub(crate) mod hazmat;
 
 mod encdec;
 mod expand;
+mod intrinsics;
 #[cfg(test)]
 mod test_expand;
 
--- a/aes/src/armv8/encdec.rs
+++ b/aes/src/armv8/encdec.rs
@@ -4,6 +4,12 @@ use crate::{Block, Block8};
 use cipher::inout::InOut;
 use core::arch::aarch64::*;
 
+// Stable "polyfills" for unstable core::arch::aarch64 intrinsics
+// TODO(tarcieri): remove when these intrinsics have been stabilized
+use super::intrinsics::{
+    vaesdq_u8, vaesdq_u8_and_vaesimcq_u8, vaeseq_u8, vaeseq_u8_and_vaesmcq_u8,
+};
+
 /// Perform AES encryption using the given expanded keys.
 #[target_feature(enable = "aes")]
 #[target_feature(enable = "neon")]
@@ -19,11 +25,8 @@ pub(super) unsafe fn encrypt1<const N: usize>(
     let mut state = vld1q_u8(in_ptr as *const u8);
 
     for k in expanded_keys.iter().take(rounds - 1) {
-        // AES single round encryption
-        state = vaeseq_u8(state, *k);
-
-        // AES mix columns
-        state = vaesmcq_u8(state);
+        // AES single round encryption and mix columns
+        state = vaeseq_u8_and_vaesmcq_u8(state, *k);
     }
 
     // AES single round encryption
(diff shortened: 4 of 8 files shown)


tarcieri:
Oh hmm, the target features were stabilized sometime after 1.59. Guess I'll have to figure out when.

tarcieri:
Weird, it wouldn't compile on `aarch64-unknown-linux-gnu` without 10debe5 which causes a performance regression.

Let me play around with it a bit more. It's working fine on `aarch64-apple-darwin` and I would think `#[inline(always)]` would work in place of `target_feature`.

newpavlov:
>Weird, it wouldn't compile on aarch64-unknown-linux-gnu without https://github.com/RustCrypto/block-ciphers/commit/10debe547fdae5612fdf72fb445ad5ed4904b27e which causes a performance regression.

Huh, it's certainly looks weird. I don't think I've seen such behavior on x86. I would think that asm blocks should be completely opaque for compiler (apart from register allocation stuff). Maybe ask about it on IRLO or in a new Rust repository issue?

tarcieri:
Yeah, seems like some sort of bug in rustc. You can see the build failure here:

https://github.com/RustCrypto/block-ciphers/actions/runs/5283017067/jobs/9558698657

Curiously it only seems to care about `AESE` and `AESIMC`. It doesn't seem to care about `AESD` and `AESMC`.

tarcieri:
I'm pretty perplexed by the performance difference of https://github.com/RustCrypto/block-ciphers/commit/10debe547fdae5612fdf72fb445ad5ed4904b27e too... it seems like unrelated code makes the performance greatly suffer

```
$ RUSTFLAGS="--cfg aes_armv8" cargo +nightly bench
   Compiling aes v0.8.2 (/Users/tony/src/RustCrypto/block-ciphers/aes)
    Finished bench [optimized] target(s) in 1.40s
     Running unittests src/lib.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/aes-eb55ebb5f7ca4ca0)

running 4 tests
test armv8::test_expand::aes128_key_expansion ... ignored
test armv8::test_expand::aes128_key_expansion_inv ... ignored
test armv8::test_expand::aes192_key_expansion ... ignored
test armv8::test_expand::aes256_key_expansion ... ignored

test result: ok. 0 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/mod.rs (/Users/tony/src/RustCrypto/block-ciphers/target/release/deps/mod-6305260e7c7483e7)

running 15 tests
test aes128_decrypt_block  ... bench:       1,060 ns/iter (+/- 9) = 15456 MB/s
test aes128_decrypt_blocks ... bench:       1,027 ns/iter (+/- 14) = 15953 MB/s
test aes128_encrypt_block  ... bench:       2,698 ns/iter (+/- 43) = 6072 MB/s
test aes128_encrypt_blocks ... bench:       3,606 ns/iter (+/- 170) = 4543 MB/s
test aes128_new            ... bench:         104 ns/iter (+/- 4)
test aes192_decrypt_block  ... bench:       1,447 ns/iter (+/- 14) = 11322 MB/s
test aes192_decrypt_blocks ... bench:       1,273 ns/iter (+/- 11) = 12870 MB/s
test aes192_encrypt_block  ... bench:       3,032 ns/iter (+/- 93) = 5403 MB/s
test aes192_encrypt_blocks ... bench:       3,607 ns/iter (+/- 172) = 4542 MB/s
test aes192_new            ... bench:         103 ns/iter (+/- 3)
test aes256_decrypt_block  ... bench:       1,755 ns/iter (+/- 85) = 9335 MB/s
test aes256_decrypt_blocks ... bench:       1,571 ns/iter (+/- 28) = 10429 MB/s
test aes256_encrypt_block  ... bench:       3,417 ns/iter (+/- 22) = 4794 MB/s
test aes256_encrypt_blocks ... bench:       4,148 ns/iter (+/- 268) = 3949 MB/s
test aes256_new            ... bench:         124 ns/iter (+/- 1)

test result: ok. 0 passed; 0 failed; 0 ignored; 15 measured; 0 filtered out; finished in 22.36s
```

newpavlov:
My guess is that it fails to inline the "intrinsics", despite having `#[target_feature(enable = "aes")]` on their users. It could be worth to explore code generation in godbolt to find what exactly causes codegen degradation.

tarcieri:
@newpavlov what's really weird is that the performance of `vaesdq_u8_and_vaesimcq_u8` is massively degraded, but it doesn't call either `vaeseq_u8` or `vaesimcq_u8`, it has its own separate block of assembly.

Replacing `#[inline(always)]` with `#[target_feature(enable = "aes")]` is causing a performance degradation on seemingly unrelated code. Perhaps you're right that it's breaking inlining somehow, but the functions are only related by their contents (i.e. the same ASM instructions), not a direct call path.

tarcieri:
I'm also having trouble extracting a minimal repro. Just `encrypt1`/`encrypt8`/`decrypt1`/`decrypt8` and the intrinsics seem to be fine, so I guess it's something with the key expansion.

tarcieri:
I opened a rustc issue here: https://github.com/rust-lang/rust/issues/112709

tarcieri:
Okay, it seems there were some callers in the key expansion code which weren't properly annotated with `target_feature(enable = "aes")` which were the culprit: https://github.com/RustCrypto/block-ciphers/pull/365/commits/c7fe62edcdb121b709af1689dda1214f435007af

The OS-specific discrepancies were due to `target_feature=+aes` being enabled on certain targets but not others, with the error occurring where it wasn't enabled.

With that new commit we're green on all targets, and I've confirmed performance is not impacted, so this is ready for review.

tarcieri:
@newpavlov hmm, perhaps we should use `#[inline]` + `#[target_feature]` instead of `#[inline(always)]` to avoid these hard-to-debug errors: https://github.com/rust-lang/stdarch/pull/306

aes/src/armv8/expand.rs
```diff
@@ -37,9 +41,9 @@ pub(super) fn expand_key<const L: usize, const N: usize>(key: &[u8; L]) -> [uint
         let mut word = ek_words[i - 1];
 
         if i % nk == 0 {
-            word = sub_word(word).rotate_right(8) ^ ROUND_CONSTS[i / nk - 1];
+            word = unsafe { sub_word(word) }.rotate_right(8) ^ ROUND_CONSTS[i / nk - 1];
         } else if nk > 6 && i % nk == 4 {
-            word = sub_word(word)
+            word = unsafe { sub_word(word) };
```
taiki-e, line 46: Since `expand_key` is not marked as `#[target_feature(enable = "aes")]`, I think these `sub_word` calls would not be inlined if `aes` target feature is not enabled at compile time. I wonder if that may potentially affect performance.

tarcieri: Yeah, I could just throw `#[target_feature(enable = "aes")]` on it and see how it impacts performance.

The performance is already something on the order of something like 80ns though, and this is just the initial key setup, so it's something of a micro-optimization. I can see what the impact is.

tarcieri: I tried it and it was 64ns with or without (noise of +/-2). It's a change I might make if I refactor that method anyway though.

tarcieri:
Hmm, interesting, on an M2 Max the "parallel" versions seem to offer no speedup and in fact perform worse:

```
test aes128_decrypt_block  ... bench:         838 ns/iter (+/- 24) = 19551 MB/s
test aes128_decrypt_blocks ... bench:         851 ns/iter (+/- 9) = 19252 MB/s
test aes128_encrypt_block  ... bench:         845 ns/iter (+/- 46) = 19389 MB/s
test aes128_encrypt_blocks ... bench:         853 ns/iter (+/- 26) = 19207 MB/s
test aes128_new            ... bench:          64 ns/iter (+/- 4)
test aes192_decrypt_block  ... bench:         984 ns/iter (+/- 27) = 16650 MB/s
test aes192_decrypt_blocks ... bench:       1,004 ns/iter (+/- 46) = 16318 MB/s
test aes192_encrypt_block  ... bench:       1,002 ns/iter (+/- 65) = 16351 MB/s
test aes192_encrypt_blocks ... bench:       1,041 ns/iter (+/- 90) = 15738 MB/s
test aes192_new            ... bench:          62 ns/iter (+/- 3)
test aes256_decrypt_block  ... bench:       1,164 ns/iter (+/- 31) = 14075 MB/s
test aes256_decrypt_blocks ... bench:       1,349 ns/iter (+/- 105) = 12145 MB/s
test aes256_encrypt_block  ... bench:       1,180 ns/iter (+/- 186) = 13884 MB/s
test aes256_encrypt_blocks ... bench:       1,340 ns/iter (+/- 69) = 12226 MB/s
test aes256_new            ... bench:          84 ns/iter (+/- 5)
```

aes/src/armv8/expand.rs
taiki-e: On which platform did you benchmark?

On aarch64 macOS, `aes` target feature is enabled at compile time by default so the compiler can inline all `#[target_feature(enable = "aes")]` functions regardless of whether `#[target_feature(enable = "aes")]` is used by the caller. The aes feature must be disabled at compile time to measure performance regarding the optimization I'm discussing.


In any case, I agree with your opinion that this should not significantly improve performance, because it is not called in a hot path.

tarcieri: Aah yes, it was on macOS

newpavlov:
@tarcieri 
Such micro-benchmarks are quite finicky, especially on sub-microsecond scales. The difference is within the measurement error, so I would say the performance is effectively the same.

Also note that the best number of blocks processed in parallel is processor-dependent. It depends not only on latency/throughput of used instructions, but also on number of available registers. So even within the same target arch it may vary. 8 blocks is the optimal number for mainstream x86 processors, while on certain older x86 CPUs 6 blocks would perform better. So on M1/M2 a different number could produce better results, or maybe we even don't need explicit parallel processing for it, since these CPUs are famously wide with deep reordering buffers.

tarcieri:
@newpavlov for AES-256 in particular it's outside the noise threshold, where the noise floor would be around 1,271 ns and the non-parallel version is around 1,180 ns. It's also fairly consistent and reproducible.

I added the parallelism because there was a noticeable benefit on earlier M1 CPUs however, which are much more widespread.

Anyway, just an observation.

newpavlov:
Maybe try to experiment with different numbers of parallel blocks? It would be interesting to see a plot with such data for M1 and M2.

Also, maybe there are Apple recommendations for implementing AES similar to the Intel ones?

newpavlov (approved):

PR #379, opened 2023-08-22
bump hex-literal for kuznyechik to 0.4

alexanderkjall:
ref: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/kuznyechik/debian/patches/relax-deps.patch


newpavlov:
https://github.com/RustCrypto/block-modes/pull/50#issuecomment-1688500814

PR #384, opened 2023-09-15
cast6: initial crate

LegionMammal978:
This crate is mostly a copy-paste of the `cast5` code. Apart from everything being larger in size, there are a couple main differences: `forward_quad()` and `backward_quad()` replace the direct `fX!()` calls, and `Cast6::key_schedule()` is implemented more regularly in terms of `forward_octave()`. The `TM` and `TR` arrays were generated from the algorithm in the RFC; since it only takes a bunch of wrapping additions, it might be worthwhile to construct one or both arrays with a const function, or even to compute each element at runtime.


tarcieri:
Going to merge this in the interest of getting it in before we do a `cipher` upgrade across the whole repo.

@newpavlov can you grant crate publishing access to the block ciphers team?

newpavlov:
@tarcieri Done.

PR #390, opened 2023-11-03
sm4: simd support for armv8, x86_64

zonyitoo:
- x86_64: aesni, avx2
- armv8: crypto-extension, neon


newpavlov:
Thank you! It's a relatively sizable PR, so it will take some time for us to properly review it.

zonyitoo:
Hello @newpavlov , what do you think about this PR?

sm4/src/x86/aesni.rs
```diff
@@ -0,0 +1,374 @@
+//! SM4 with X86 AES-NI instruction set
+//!
+//! Implementation was borrowed from <https://www.cnblogs.com/kentle/p/15826075.html> by kentle.
```
tarcieri, line 3: I found an English description of the approach being used here which helped me understand how this actually works:

https://github.com/mjosaarinen/sm4ni

> The trick is to use affine transforms to emulate the SM4 S-Box with the AES S-Box. The S-Boxes are both based on finite field inversion, but use different affine transforms and even polynomial basis for the finite field. However, different polynomial bases are affine isomorphic.

Interesting approach!

tarcieri:
@zonyitoo it looks pretty impressive but as @newpavlov said it is large and because of that hard to review, especially because it contains so many backends in a single PR.

We're about to start making some breaking changes and I was looking through to see if we should land this PR first. I think it would probably make sense to try to land this as part of our next breaking release cycle. See #394

sm4/src/x86/aesni.rs
zonyitoo: Yes it is.

zonyitoo:
Sorry for the very very long reply since the last review. I was waiting for the final release of `cipher v0.5.0`, but it was still in pre-release state until now.

ShenFeng312:
good work !  I need this!

zonyitoo:
@tarcieri Hello, do you have time to review this PR?

zonyitoo:
Just found that x86 also has SM4 support in its latest CPUs, but rustc's `sm4` feature is still unstable.

newpavlov:
@zonyitoo 
The SM4 instructions are supported in inline assembly, so you could use `asm!` to create temporary polyfills for the missing intrinsics. Annoyingly, you also technically should enable the `sm4` target feature even when the instructions are used through inline assembly (right now it works fine in practice without it, but it may break later), so you still have to rely on Nightly for it.

zonyitoo:
> @zonyitoo The SM4 instructions are supported in inline assembly, so you could use `asm!` to create temporary polyfills for the missing intrinsics. Annoyingly, you also technically should enable the `sm4` target feature even when the instructions are used through inline assembly (right now it works fine in practice without it, but it may break later), so you still have to rely on Nightly for it.

Yes, I just did it in my local branch. But since this PR is already too large, I would make another one after this PR is merged.

> SM4 is supported by Intel processors, starting from [Arrow Lake S](https://en.wikipedia.org/wiki/Arrow_Lake_(microprocessor)), [Lunar Lake](https://en.wikipedia.org/wiki/Lunar_Lake), Diamond Rapids and Clearwater Forest.[[12]](https://en.wikipedia.org/wiki/SM4_(cipher)#cite_note-12)

which are rather new cpu models.

zonyitoo:
Merged latest master changes.

PR #394, opened 2023-12-30
Bump `cipher` to v0.5.0-pre.1; MSRV 1.65

tarcieri:
The main notable change is the replacement of `generic-array` with `hybrid-array`, which brings with it an MSRV of 1.65.


tarcieri:
Relatively straightforward upgrade I think. Mostly find/replace.

We might consider a crate release of `cipher` so the `minimal-versions` checks pass.

newpavlov (approved):

PR #401, opened 2024-01-22
`belt-block` expose `from_u32` to pub

makavity:
Because AEAD uses this to get key, expose it


newpavlov:
I think this function is small enough and can be easily copied. It feels somewhat wrong to expose such function from a block cipher crate.

makavity:
Sounds logically, so `to_u32` in crate used only to format key?

newpavlov:
Ah, I've missed that the crate already exports `to_u32`. We probably should make it private in the next breaking release.

For now, I would recommend to copy `to_u32` and `from_u32` to downstream crates. It may be even worth to try add more generic methods to `std`, we could use it in other crates as well.

makavity:
So, okay. As here is migration for 0.2, i suggest to make to_u32 private also

newpavlov:
The `to_u32` function was made private in #402.

PR #403, opened 2024-01-26
Implement AES backend for ARMv9 using SVE2-AES

silvanshade:
This PR implements an AES backend for ARMv9 using the SVE2-AES feature.

The overall design is similar to [RVV implementation](https://github.com/RustCrypto/block-ciphers/pull/399) since they are both VLA-style.

Most of the comments I made in that PR are relevant here as well, especially with regard to the use of `global_asm!` and the opportunity for an addition to the interface for supporting N-ary parallel operations (although maybe this doesn't need to be a user-facing change, depending on the specifics).

One thing to note for anyone checking the CI logs: there's unfortunately an annoying issue that causes some part of the LLVM code-gen to complain about assembly using extensions (when compiled in release mode), indicating there is an error in compilation, although the code actually compiles and runs fine and all the tests pass.

More details with link to an upstream issue: https://github.com/rust-lang/rust/issues/80608

In fact this was also a problem for the RVV implementation but I was able to use the RISC-V assembly `.architecture` directive to silence the errors, like so:

```asm
.attribute arch, "rv64gcv1p0_zkne_zknd_zvkned1p0"
```

Unfortunately, ARMv9 assembly doesn't have an analogous directive, AFAIK. However, something could be added to the job to filter out the noise from these spurious notices if desired.

Also relevant (for switching to intrinsics in the future): https://github.com/rust-lang/rust/pull/118917

### DIFF
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -16,6 +16,7 @@ defaults:
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
+  QEMU_SRC_VERSION: 8.2.0
 
 jobs:
   # Builds for no_std platforms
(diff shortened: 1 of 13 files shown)


tarcieri:
Seems like a lot of duplication between the ARMv8 and ARMv9 backends. Perhaps you could consolidate them? Maybe a module structure like:

- `arm`
  - `v8`
  - `v9`

...would make sense?

silvanshade:
I can give it a try, sure. Will be a few days probably since I need to finish up some other project first.

The main difficulty for me with this one (and the RVV one which I can also look at again) is lack of access to hardware for real-world benchmarks. The ARM9 SVE implementation we should be able to test on EC2 Graviton I think. Maybe there are other options.

PR #406, opened 2024-02-10
Fix kuznyechik 32bit build

silvanshade:


newpavlov:
Thank you! We probably need to do a backport for v0.8, since `master` currently tracks future v0.9 release.

PR #407, opened 2024-02-13
aes: Assert precondition required for soundness

joshlf:
Feel free to edit, and let me know if I've missed something and this is actually guaranteed statically.

This came up [here](https://fuchsia-review.git.corp.google.com/c/fuchsia/+/987054/1/third_party/rust_crates/vendor/aes-0.8.3/src/armv8/expand.rs#26) when vendoring `aes` into Fuchsia.

### DIFF
--- a/aes/src/armv8/expand.rs
+++ b/aes/src/armv8/expand.rs
@@ -18,6 +18,9 @@ pub unsafe fn expand_key<const L: usize, const N: usize>(key: &[u8; L]) -> [uint
 
     let mut expanded_keys: [uint8x16_t; N] = mem::zeroed();
 
+    // Sanity check, as this is required in order for the following line to be sound.
+    // We'd rather panic than exhibit undefined behavior.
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks. I've tried to refactor this code a few times but never found an approach that felt like an improvement.

erickt:
Thanks @tarcieri! Would it be possible to cut a 0.8.4 release with it?

newpavlov:
Maybe we should do it using const asserts? Something like this:
```rust
const _: () = {
    assert!(mem::align_of::<uint8x16_t>() >= mem::align_of::<u32>());
};
```

>Would it be possible to cut a 0.8.4 release with it?

We would need to do a backport. `master` currently tracks future v0.9 releases.

tarcieri:
I can do a backport if need be. Also merged @newpavlov's const assert PR.

erickt:
Thanks so much!

tarcieri:
Released in `aes` v0.8.4

PR #411, opened 2024-02-20
Bump `cipher` dependency to `0.5.0-pre.3`

baloo:

PR #413, opened 2024-03-04
bump `cipher` dependency to `0.5.0-pre.4`

baloo:
supersedes #411


baloo:
relaxed the dependency to cipher, I previously believed we needed that, but apparently it's not required.
(context: https://github.com/RustCrypto/universal-hashes/pull/196#discussion_r1510414327) I believe this one is suitable for a more relaxed version.

tarcieri (approved):

PR #428, opened 2024-07-22
Include Apache license text in rc2 and rc5 crates

davide125:
The `rc2` and `rc5` crates do not currently include the full license text, which is technically required by the Apache license.


newpavlov:
Thank you!

PR #431, opened 2024-07-26
Cut pre.1 prereleases

tarcieri:
Cuts a new release of every crate which previously received a pre.0 prerelease, i.e. is used as a (dev-)dependency in downstream projects that need to be upgraded.

This includes the following:

- `aes` v0.9.0-pre.1
- `belt-block` v0.2.0-pre.1
- `des` v0.9.0-pre.1
- `kuznyechik` v0.9.0-pre.1
- `magma` v0.10.0-pre.1

### DIFF
--- a/aes/Cargo.toml
+++ b/aes/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aes"
-version = "0.9.0-pre.0"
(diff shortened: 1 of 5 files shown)


tarcieri:
I was able to publish everything except `belt-block`

PR #435, opened 2024-07-29
Fix new Clippy lints

newpavlov:


aes/src/ni/aes192.rs
```diff
@@ -115,10 +115,11 @@ macro_rules! expand_round {
...
 
-macro_rules! shuffle {
-    ($a:expr, $b:expr, $imm:expr) => {
-        mem::transmute::<_, __m128i>(_mm_shuffle_pd(mem::transmute($a), mem::transmute($b), $imm))
-    };
+#[inline(always)]
+unsafe fn shuffle(a: __m128i, b: __m128i, i: usize) -> __m128i {
+    let a: [u64; 2] = mem::transmute(a);
+    let b: [u64; 2] = mem::transmute(b);
+    mem::transmute([a[i], b[0]])
```
newpavlov, line 122: This function conveys the intent better and gets properly compiled into `shufpd` and `unpcklpd` instructions after inlining.

PR #436, opened 2024-07-29
gift: use built-in benchmarks instead of Criterion

newpavlov:
Criterion is a relatively big dependency which unnecessarily gets pulled during testing. Additionally, it results in the [security warning](https://github.com/RustCrypto/block-ciphers/security/dependabot/1).

In addition to the benchmark changes, this PR also slightly tweaks module structure of the crate.


.cargo/audit.toml
```diff
@@ -1,7 +0,0 @@
-# https://github.com/RustSec/cargo-audit/blob/master/audit.toml.example
-
-[advisories]
-ignore = [
-    "RUSTSEC-2021-0127", # serde_cbor
-    "RUSTSEC-2021-0145", # atty
```
newpavlov, line 6: These dependencies were pulled by the old Criterion version, so we no longer need these exceptions.

PR #442, opened 2024-07-31
aes: rework backends

newpavlov:
This PR unifies code between AES-NI and ARM backends and prepares ground for future removal of duplicated definitions of AES types in `autodetect`, `soft`, `ni`, and `armv8` modules. Additionally, it allows to quickly change number of blocks processed in parallel by different intrinsics-based backends instead of hardcoding it to 8 blocks.


aes/src/macros.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        impl cipher::KeyInit for $enc_name {
+            #[inline]
+            fn new(key: &cipher::Key<Self>) -> Self {
+                let keys = unsafe { $expand_keys(key.as_ref()) };
+                Self { keys }
+            }
+        }
+
+        impl cipher::BlockBackend for &$enc_name {
```
newpavlov, line 42: `BlockBackend` is implemented for reference because its methods works with `&mut self`. We probably should introduce two separate traits `BlockCipherBackend` (with `&self` methods) and `BlockModeBackend` (with `&mut self` methods).

aes/src/ni.rs
```diff
@@ -33,23 +28,62 @@ use core::arch::x86 as arch;
...
+    BlockSizeUser, Key, KeyInit, KeySizeUser,
 };
 use core::fmt;
 
+impl_backends!(
+    enc_name = Aes128BackEnc,
+    dec_name = Aes128BackDec,
+    key_size = consts::U16,
+    keys_ty = expand::Aes128RoundKeys,
+    par_size = consts::U15,
```
newpavlov, line 43: Since x86 has only 16 XMM registers (AVX-512 is out of scope for now), processing 15 blocks in parallel on x86 means that each round key will be loaded on each iteration. It maximizes ILP, but introduces additional loads from L1 cache.

For AES-128, 192, and 256 we can process only 5, 3, and 1 block in parallel respectively without reloading some keys.  On my laptop the sweet spot seems to be 11 blocks (according to the crate benchmarks), but it's likely highly dependent on CPU model. We will need additional benchmarks including the CTR mode to find optimal numbers. For now I decided to use 15 blocks for assembler simplicity.

aes/src/armv8.rs
```diff
@@ -17,20 +17,55 @@ mod expand;
...
 };
-use core::arch::aarch64::*;
 use core::fmt;
 
+impl_backends!(
+    enc_name = Aes128BackEnc,
+    dec_name = Aes128BackDec,
+    key_size = consts::U16,
+    keys_ty = expand::Aes128RoundKeys,
+    par_size = consts::U15,
```
newpavlov, line 32: ARMv8 NEON has 32 SIMD registers, so techincally we can process 21, 19, and 17 blocks in parallel for AES-128, 192, and 256 respectively while keeping round keys in registers. But since the code forces inlining, it will balloon binary size, so additional benchmarks are needed.

PR #445, opened 2024-08-06
CI: bump actions/checkout to v4 and trigger workflows on config changes

newpavlov:


tarcieri:
Maybe we should configure Dependabot to open these automatically

newpavlov:
Checkout updates look rare enough, so I am fine with either.

tarcieri:
Dependabot would be nice for automatically updating all of the repos, though

PR #451, opened 2024-09-26
kuznyechik: use const eval to generate tables

### LINKED ISSUE #448: kuznyechik: use `const fn` to compute the fused tables
Right now the tables are included as binary blobs which not only make it harder to review them, but also increase size of the crate. Ideally, the tables should be computed at compile time based on the S-box constant.

newpavlov:
This PR significantly reduces size of the crate. But now clean build of the crate takes 7-12 seconds, which is not great, but tolerable enough. Hopefully, performance of const eval will be improved in future.

We could improve build times by using a build script and `include_bytes!`, but it would result in a somewhat less clear code and it would mean introduction of the first algorithmic crate which uses build script.

Additionally, this PR adds a chain encryption test for a more thorough testing.

Closes #448


kuznyechik/src/fused_tables.rs
```diff
@@ -0,0 +1,104 @@
...
+                x ^= GFT_192[table[pos + 10] as usize];
+                x ^= GFT_194[table[pos + 11] as usize];
+                x ^= GFT_16[table[pos + 12] as usize];
+                x ^= GFT_133[table[pos + 13] as usize];
+                x ^= GFT_32[table[pos + 14] as usize];
+                x ^= GFT_148[table[pos + 15] as usize];
+
+                let mut k = 0;
+                while k < 15 {
+                    table[pos + k] = table[pos + k + 1];
```
newpavlov, line 90: We can eliminate these copies using the same approach as in the `l_step` function.

newpavlov: Using the `l_step` function unfortunately results in code exceeding the code eval limit.

PR #463, opened 2025-01-28
aes: move `Block8` to the hazmat module

### LINKED ISSUE #462: aes 0.9.0-pre.2 uses ParBlocksSize of 9 but has type alias for Block8
I recently used the current pre-release version of the aes crate and stumbled on a performance pitfall. In previous versions, the AES-NI implementation processed 8 blocks in parallel. This was changed in #442 to 9 blocks (see [this comment](https://github.com/RustCrypto/block-ciphers/pull/442#discussion_r1698903449)). I chunked an iterator into 8 block chunks but got no speedup due to this.

Unfortunately, aes still exposes the [`Block8`](https://docs.rs/aes/0.9.0-pre.2/aes/type.Block8.html) type alias. Personally, I thought the reason for the type alias was it's usefulness when wanting to make use of parallel processing. Admittedly, this is not described by the `Block8` documentation but supported by the [hazmat API](https://docs.rs/aes/0.8.4/aes/hazmat/fn.cipher_round_par.html).

So my first suggestion would be to expand the docs of Block8 to clarify that its existence is not tied to the number of parallel rounds (which depend on the backend anyway). If this type is useful only in the context of the hazmat API, maybe it could even be moved there before the release of 0.9.

While I'm on the topic of ParBlockSize: Is there currently any way to get the ParBlocksSize for a Backend? If not, should I create an issue for this?

newpavlov:
The backends were changed to use a different number of parallel blocks, so the alias is now used only by hazmat functions.

Closes #462

### DIFF
--- a/aes/src/armv8/hazmat.rs
+++ b/aes/src/armv8/hazmat.rs
@@ -4,7 +4,7 @@
 //! implementations in this crate, but instead provides raw AES-NI accelerated
 //! access to the AES round function gated under the `hazmat` crate feature.
 
-use crate::{Block, Block8};
+use crate::hazmat::{Block, Block8};
 use core::arch::aarch64::*;
 
 /// AES cipher (encrypt) round function.
--- a/aes/src/hazmat.rs
+++ b/aes/src/hazmat.rs
@@ -11,7 +11,11 @@
 //! We do NOT recommend using it to implement any algorithm which has not
 //! received extensive peer review by cryptographers.
 
-use crate::{soft::fixslice::hazmat as soft, Block, Block8};
+use crate::soft::fixslice::hazmat as soft;
+
+pub use crate::Block;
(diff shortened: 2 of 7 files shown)


tarcieri (commented):
I guess I'm confused why the non-hazmat APIs use 9 blocks instead of 8, but this otherwise seems fine

newpavlov:
See UPD to this [comment](https://github.com/RustCrypto/block-ciphers/pull/442#discussion_r1698903449). It has shown a better performance on several (AMD) CPUs which I've tested the code.

tarcieri:
It would probably be good to test the impact on full AEADs. Seems like having an odd number of blocks is going to make optimizing parallel pipelining (which we don't do, because it makes things slower already) that much harder: https://github.com/RustCrypto/AEADs/issues/74

PR #468, opened 2025-02-12
des: tweak weak key test

newpavlov:
The new code results in a slightly better codegen while still being const time, including 32 bit targets: https://rust.godbolt.org/z/K7cjj5d1P


newpavlov:
@baloo 
I am not sure what exact purpose of the bit counting loop was, so I removed it for now. Could you explain it?

baloo:
That was a copy paste from https://github.com/iqlusioninc/yubikey.rs/blob/13bdf9a5850c124057825297b51eaecaf8ebe9e2/src/mgm/tdes.rs#L36
I think @str4d was the original author.

newpavlov:
I added subkey equality check for TDES in #470.

PR #472, opened 2025-02-23
Upgrade all crates to the 2024 edition and bump MSRV to 1.85

baloo:


aes/src/armv8/encdec.rs
```diff
@@ -18,21 +18,23 @@ pub(super) unsafe fn encrypt<const KEYS: usize>(
...
+    let mut block = unsafe { vld1q_u8(in_ptr.cast()) };
 
     for &key in &keys[..KEYS - 2] {
-        // AES single round encryption
-        block = vaeseq_u8(block, key);
-        // Mix columns
-        block = vaesmcq_u8(block);
+        unsafe {
+            // AES single round encryption
+            block = vaeseq_u8(block, key);
```
newpavlov, line 26: I think it's better to just enable the lint for such modules than to introduce this amount of `unsafe` blocks.

newpavlov (commented):
You also need to update `hex-literal` to v1.

aes/src/lib.rs
```diff
@@ -115,6 +115,7 @@
 )]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![warn(missing_docs, rust_2018_idioms)]
+#![allow(unsafe_op_in_unsafe_fn)]
```
newpavlov, line 118: I think it's better to do this `allow` on the per-module basis like I did in the hashes PR. With potential `target_feature` improvements (which would make some intrinsics safe inside functions with with enabled target features) we may enable this lint back.

baloo: Ha, I've tried to read the patch for hashes, but github UI is very slow. I've thought you've used a big `unsafe{` block, but that was only in `sha1`, I should have kept reading, my bad.

newpavlov: I did it in the initial commits, but then switched to `#[allow(..)]`. For example [here](https://github.com/RustCrypto/hashes/blob/cfe82a0f1a9a6c89c5b6d8309f2603c644dc97ea/sha1/src/compress/x86.rs#L34) I enable it for the function and [here](https://github.com/RustCrypto/hashes/blob/cfe82a0f1a9a6c89c5b6d8309f2603c644dc97ea/sha2/src/sha256/aarch64.rs) for the whole module.

newpavlov (approved):

PR #474, opened 2025-02-25
serpent: Actually bitslice encryption and decryption

randombit:
The Serpent sboxes were designed to be bitsliced, but previously were implemented via table lookups instead.

This change improves performance by approximately a factor of 20. In addition, the avoidance of table lookups removes the risk of cache-based side channels.


randombit:
With the sbox bitsliced it becomes quite trivial to execute in parallel using SIMD instructions, and Rust should easily be able to handle this own its own with just some code restructuring to make the parallelism visible. I am trying to keep this PR as small as practical, so this just introduces the bitslice sbox expressions without making any major changes to the implementation otherwise.

serpent/src/consts.rs
```diff
@@ -1,25 +1,2 @@
 pub const PHI: u32 = 0x9e37_79b9;
 pub const ROUNDS: usize = 32;
```
randombit, line 2: `consts.rs` is pretty bare at this point, I wasn't sure if it should be kept as is or just moved to `lib.rs`

newpavlov: Yes, we can move these two constants to `lib.rs`.

randombit:
This approach goes back to [the original design paper](https://www.cl.cam.ac.uk/archive/rja14/Papers/serpent.pdf) though for whatever reason explicit expressions for the Sboxes are not included there

> Our cipher has therefore been designed so that all operations can be executed using 32-fold parallelism during the encryption or decryption of a single block.

> S-Boxes: The 128-bit combination of input and key is considered as four 32-bit words. The S-box, which is implemented as a sequence of logical operations (as it would be in hardware) is applied to these four words, and the result is four output words. The CPU is thus employed to execute the 32 copies of the S-box simultaneously

The [AES submission software](https://www.cl.cam.ac.uk/archive/rja14/serpent.html) implements Serpent in this way, though with relatively inefficient expressions. The specific bitslice formulas here are from [Speeding Up Serpent](https://www.ii.uib.no/~osvik/pub/aes3.pdf) which are optimized for a register-poor 2-operand instruction set.

newpavlov:
>The specific bitslice formulas here are from [Speeding Up Serpent](https://www.ii.uib.no/~osvik/pub/aes3.pdf) which are optimized for a register-poor 2-operand instruction set.

This link would fit perfectly in the module comments.

Do you know about alternatives for less register-constrained arches like ARM and RISC-V? I think the current implementation could also be easily ported to SSE and AVX. But I guess we can leave it for later PRs.

randombit:
> Do you know about alternatives for less register-constrained arches like ARM and RISC-V?

I'm not aware of any recent searches along these lines. The search that found these expressions was, IIUC, designed to support a 2-issue superscalar processor. Likely these days for almost all processors, expressions designed to execute 3 or  4 in parallel would be better, even if they used more registers and/or gates. Most instruction sets these days also support some kind of negated `and` (either `~(a & b)` or `~a & b`, or both) which was not considered at the time.

> I think the current implementation could also be easily ported to SSE and AVX. 

Yes it extends very naturally to SSE2/AVX2/AVX-512/NEON/etc. Using AVX2 speeds in excess of 500 MB/s are easily achieved, as long as the mode supports parallel operation. In AVX-512 one can even [take advantage of `vpternlogd`](https://github.com/randombit/botan/blob/master/src/lib/block/serpent/serpent_avx512/serpent_avx512.cpp) since that provides all 256 possible 3-operand bitwise functions.

newpavlov:
BTW I tried to fully unroll the rounds loop, but strangely the performance has dropped on my laptop from ~38 MB/s to ~33 MB/s. I wonder if it's something related to the scheduling issues mentioned in the paper.

randombit:
In 32ef6b1 I added `#[inline]` annotations and at least on my machine (AMD Zen2 laptop) performance went from 33 MB/s to 111 MB/s. TBH I'm a little surprised that `rustc` wasn't inlining these functions already without the hint; most are on the order of 20 instructions so inlining seems like an easy win to me. I'll have to keep that in mind for the future.

On the same machine, with their relevant SIMD support disabled, the C++ implementation in Botan and the C implementation in libgcrypt both run around 110-120 MB/s. So what I'm seeing now seems about right for a scalar implementation of Serpent. I had been, up until adding `#[inline]` annotations, confused why this version (with bitslicing) was still ~ a quarter the speed I would have expected.

> BTW I tried to fully unroll the rounds loop, but strangely the encryption performance on my AMD laptop has dropped from ~38 MB/s to ~33 MB/s.

It may have been an artifact of the relevant operations (linear transform + sbox) not being inlined; if `rustc` wasn't inlining these, then I guess unrolling the loop wouldn't help much; instead of a 32 iteration loop with 2 function calls in the body, you got 64 function calls in a straight line. Either way there's a ton of overhead for function call setup and teardown, and LLVM can't interleave the operations to improve scheduling or register usage. [It does still seem weird that performance went _down_ rather than just being flat...]

serpent/src/lib.rs
```diff
@@ -34,21 +34,384 @@ use core::fmt;
 use cipher::zeroize::{Zeroize, ZeroizeOnDrop};
 
 mod consts;
-use consts::{PHI, ROUNDS, S, S_INVERSE};
+use consts::{PHI, ROUNDS};
 
 type Words = [u32; 4];
 type RoundKeys = [Words; ROUNDS + 1];
 
+const fn sbox_e0([mut w1, mut w2, mut w3, mut w4]: Words) -> Words {
```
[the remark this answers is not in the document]
randombit: I moved these to `bitslice.rs`

serpent/src/bitslice.rs
```diff
@@ -0,0 +1,399 @@
+/*
+ * Serpent uses 8 4-bit Sboxes which were designed to be implemented using small
```
newpavlov, line 2: We generally use `//!` style comments.

serpent/src/lib.rs
```diff
@@ -77,49 +77,34 @@ fn linear_transform_inverse_bitslice(mut words: Words) -> Words {
...
-            get_bit(w1 as usize, i)
-                | (get_bit(w2 as usize, i) << 1)
-                | (get_bit(w3 as usize, i) << 2)
-                | (get_bit(w4 as usize, i) << 3),
-        );
-
-        for l in 0..4 {
-            words[l] |= u32::from(get_bit(quad as usize, l)) << i;
-        }
+    match index % 8 {
```
newpavlov, line 82: It's probably worth to move `apply_s(_inverse)_bitslice` and `linear_transform(_inverse)_bitslice` to the bitslice module. It would allow us to make the `sbox_*` functions private to the module.

serpent/src/bitslice.rs
randombit: :+1: changed to `//!`

serpent/src/lib.rs
randombit: Done. I also chopped `_bitslice` off the names since it seems redundant with the module name prefix.

newpavlov:
>It may have been an artifact of the relevant operations (linear transform + sbox) not being inlined; if rustc wasn't inlining these, then I guess unrolling the loop wouldn't help much; instead of a 32 iteration loop with 2 function calls in the body, you got 64 function calls in a straight line.

Hm, I tested copied snippets in godbolt and it was properly inlining the inner functions. But with your recent changes it indeed results in 100+ MB/s performance.

Unrolling the loops also helps. I did it with this helper macro:
```rust
macro_rules! repeat31 {
    ($i:ident, $body:block) => {
        let $i = 0; $body;
        let $i = 1; $body;
        let $i = 2; $body;
        let $i = 3; $body;
        let $i = 4; $body;
        let $i = 5; $body;
        let $i = 6; $body;
        let $i = 7; $body;
        let $i = 8; $body;
        let $i = 9; $body;
        let $i = 10; $body;
        let $i = 11; $body;
        let $i = 12; $body;
        let $i = 13; $body;
        let $i = 14; $body;
        let $i = 15; $body;
        let $i = 16; $body;
        let $i = 17; $body;
        let $i = 18; $body;
        let $i = 19; $body;
        let $i = 20; $body;
        let $i = 21; $body;
        let $i = 22; $body;
        let $i = 23; $body;
        let $i = 24; $body;
        let $i = 25; $body;
        let $i = 26; $body;
        let $i = 27; $body;
        let $i = 28; $body;
        let $i = 29; $body;
        let $i = 30; $body;
    };
}
```
With the round loops changed to this:
```rust
// encrypt
repeat31!(i, {
    let xb = xor(b, self.round_keys[i]);
    let s = apply_s_bitslice(i, xb);
    b = linear_transform_bitslice(s);
});

// decrypt
repeat31!(i, {
    let i = 30 - i;
    let s = linear_transform_inverse_bitslice(b);
    let xb = apply_s_inverse_bitslice(i, s);
    b = xor(xb, self.round_keys[i]);
});
```

I results in ~10% performance improvement. We can leave it for a separate PR, if you want.

randombit:
> I tested copied snippets in godbolt and it was properly inlining the inner functions. But with your recent changes it indeed results in 100+ MB/s performance, so I guess the compiler made a different decision while compiling the same code in the crate form.

Ugh, I would not have expected that. I wonder what's going on.

> interestingly, with decryption being slightly faster than encryption

I have seen this as well. My guess is that the sbox decryption circuits, while being slightly larger in gate counts, have more available ILP.

> We can leave it for a separate PR, if you want.

I'm fine with doing it here if you are. My main goal at this point is keeping the PR simple enough to review (thus for example skipping on SSE/AVX2 support for the moment, though I would like to try that in a followup)

newpavlov (approved):
Thank you!

PR #492, opened 2025-07-23
Add AES support for RISC-V: RV64 Scalar and RVV

silvanshade:
This PR adds AES support for RV64 scalar and RVV.

Some caveats:

The implementation chooses par block of 8 for scalar and 64 for RVV. I don't have any capable hardware to test this with and am not sure what the ideal configuration should be. The RVV code uses VLA style so the par block size can be easily adapted to anything. In the long run it would probably be better to provide an API that can take advantage of this directly.

If RVV is selected for the backend and scalar crypto is *not* also supported (no idea how common this is in practice), then AES-192 support is disabled. This is because RVV crypto does not provide key expansion instructions for AES-192, only for AES-128 and AES-256. If scalar crypto *is* available, we can fall back to the scalar just for key expansion for AES-192.

It would be possible to implement something special for that case instead of completely disabling it if someone wants to take that on. Maybe the fixslice implementation can be adapted somehow.

Some remaining work:

- [ ] encode tail blocks with VLA
- [ ] re-implement CI tests

### DIFF
--- a/aes/benches/mod.rs
+++ b/aes/benches/mod.rs
@@ -13,11 +13,19 @@ block_decryptor_bench!(
     aes128_decrypt_block,
     aes128_decrypt_blocks,
 );
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 block_encryptor_bench!(
     Key: aes::Aes192,
     aes192_encrypt_block,
     aes192_encrypt_blocks,
 );
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 block_decryptor_bench!(
     Key: aes::Aes192,
     aes192_decrypt_block,
@@ -43,6 +51,10 @@ fn aes128_new(bh: &mut test::Bencher) {
     });
 }
 
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 #[bench]
 fn aes192_new(bh: &mut test::Bencher) {
     bh.iter(|| {
--- a/aes/src/lib.rs
+++ b/aes/src/lib.rs
@@ -35,6 +35,26 @@
 //! runtime. On other platforms the `aes` target feature must be enabled via
 //! RUSTFLAGS.
 //!
+//! ## RISC-V rv64 (scalar) {Zkne, ZKnd} extensions
+//!
+//! Support is available for the RISC-V rv64 scalar crypto extensions for AES. This
+//! is not currently autodetected at runtime. In order to enable, you need to
+//! enable the appropriate target features at compile time. For example:
+//! `RUSTFLAGS=-C target-feature=+zkne,+zknd`.
+//!
+//! ## RISC-V rvv (vector) {Zvkned} extensions
+//!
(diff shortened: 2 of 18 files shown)


newpavlov:
Thank you! I would _really_ like for this code to be tested somehow. Do you know what status of RVV support in QEMU? IIRC the scalar crypto extension should be properly supported by it.

Also note that RISC-V has an annoyingly horrible (IMO) handling of misaligned loads, in `sha2` I had to work around this using this [hack](https://github.com/RustCrypto/hashes/blob/master/sha2/src/sha256/riscv_zknh_utils.rs). I am not familiar with RVV enough, but I heard it also has some issues with processing misaligned data.

>This is because RVV crypto does not provide key expansion instructions for AES-192, only for AES-128 and AES-256.

I think you should fall back to the soft key expansion here, we certainly should not remove `Aes192` from public exports.

newpavlov:
It also would be better to create a separate PRs for the scalar crypto support. We probably will be able to merge it faster.

silvanshade:
> I think you should fall back to the soft key expansion here, we certainly should not remove `Aes192` from public exports.

Are you suggesting to use the fixslice software implementation for key expansion? The reason I didn't use that already is because I wasn't sure how to convert from the fixslice key-expansion representation back to the standard representation.

If that's what you're suggesting, can you give me an example of how to do that?

If you're instead suggesting a new software implementation, I didn't do that because I assumed we would want a side-channel resistant implementation (like the fixslice) and wasn't sure what would be a good approach for that here.

silvanshade:
> Also note that RISC-V has an annoyingly horrible (IMO) handling of misaligned loads, in `sha2` I had to work around this using this [hack](https://github.com/RustCrypto/hashes/blob/master/sha2/src/sha256/riscv_zknh_utils.rs). I am not familiar with RVV enough, but I heard it also has some issues with processing misaligned data.

I was thinking about alignment more, trying to recall my stance on the issue when I first implemented these algorithms.

I think in my earliest implementation, I did actually perform some sort of alignment correction like what you refer to. But ultimately I decided not to include that, for a few reasons:

1. It wasn't necessary (on Linux at least)
2. It added complexity (more surface area for bugs)
3. It potentially hid issues from user with their own code

Regarding (3), it's unfortunate the user needs to care about this as much as they probably do, but I sort of feel that this is a platform hazard they ought to be aware of and that we might do more harm than good in trying to work around.

Maybe the best option here is to provide a stricter subset of the API that requires correct alignment and throws an error if the user passes in misaligned data. Then we just do our best to ensure we preserve correct alignment in our own code, which might require being more specific about it in our data structures and functions. (I don't know how much of an issue this would be for us to address).

Having said that, if you feel strongly enough that we should still try to work around alignment issues, I won't argue against that.

This assumes by the way that we aren't penalized by using calls for unaligned loads when the data is actually correctly aligned already. I haven't read through those posts of yours in detail yet. If that is in fact the case, that performance suffers just by trying to be defensive about it, then that changes things and I guess we are stuck trying to correct for it.

newpavlov:
I opened an [LLVM issue](https://github.com/llvm/llvm-project/issues/150263) with suggestion to implement the branching hack as an LLVM optimization. I am very close to just saying that it's a compiler problem and do not deal with it in our code.

silvanshade:
I added the CI tests and also split the scalar support out into a separate PR:

- https://github.com/RustCrypto/block-ciphers/pull/493

PR #493, opened 2025-07-26
Add AES support for RISC-V: RV64 scalar

silvanshade:
This PR implements AES support for RISC-V RV64 scalar crypto.

Related:

- https://github.com/RustCrypto/block-ciphers/pull/492

### DIFF
--- a/.github/workflows/aes.yml
+++ b/.github/workflows/aes.yml
@@ -310,6 +310,33 @@ jobs:
       - run: cross test --package aes --target ${{ matrix.target }} --features hazmat
       - run: cross test --package aes --target ${{ matrix.target }} --all-features
 
+  riscv:
+    strategy:
+      matrix:
+        include:
+          - target: riscv64gc-unknown-linux-gnu
+            rustflags: ''
+            rust: nightly
+          - target: riscv64gc-unknown-linux-gnu
+            rustflags: '-Ctarget-feature=+zkne,+zknd'
+            rust: nightly
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: .
+    steps:
+      - uses: actions/checkout@v4
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+      - run: cargo install cross --git https://github.com/cross-rs/cross
+        env:
+          RUSTFLAGS: "${{ env.RUSTFLAGS }} -Astatic-mut-refs"
+      - run: cross test --package aes --target ${{ matrix.target }}
+        env:
+          RUSTFLAGS: "-Dwarning ${{ matrix.rustflags }}"
+
   clippy:
     env:
       RUSTFLAGS: "-Dwarnings --cfg aes_compact"
--- a/aes/benches/mod.rs
+++ b/aes/benches/mod.rs
@@ -13,11 +13,19 @@ block_decryptor_bench!(
     aes128_decrypt_block,
     aes128_decrypt_blocks,
 );
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 block_encryptor_bench!(
     Key: aes::Aes192,
     aes192_encrypt_block,
     aes192_encrypt_blocks,
 );
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 block_decryptor_bench!(
     Key: aes::Aes192,
     aes192_decrypt_block,
@@ -43,6 +51,10 @@ fn aes128_new(bh: &mut test::Bencher) {
     });
 }
 
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
 #[bench]
 fn aes192_new(bh: &mut test::Bencher) {
     bh.iter(|| {
--- a/aes/src/lib.rs
+++ b/aes/src/lib.rs
@@ -35,6 +35,13 @@
 //! runtime. On other platforms the `aes` target feature must be enabled via
 //! RUSTFLAGS.
 //!
+//! ## RISC-V rv64 (scalar) {Zkne, ZKnd} extensions
+//!
+//! Support is available for the RISC-V rv64 scalar crypto extensions for AES. This
+//! is not currently autodetected at runtime. In order to enable, you need to
+//! enable the appropriate target features at compile time. For example:
+//! `RUSTFLAGS=-C target-feature=+zkne,+zknd`.
+//!
 //! ## `x86`/`x86_64` intrinsics (AES-NI and VAES)
 //! By default this crate uses runtime detection on `i686`/`x86_64` targets
 //! in order to determine if AES-NI and VAES are available, and if they are
@@ -123,12 +130,26 @@
 )]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![warn(missing_docs, rust_2018_idioms)]
+#![cfg_attr(
+    all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_feature = "zknd",
+        target_feature = "zkne"
+    ),
+    feature(riscv_ext_intrinsics)
+)]
 
 #[cfg(feature = "hazmat")]
 pub mod hazmat;
 
 #[macro_use]
 mod macros;
+#[cfg_attr(any(target_arch = "riscv32", target_arch = "riscv64"),
+// The `soft` AES-192 is used for RISC-V when RVV crypto (Zvkned) is supported
+// but Scalar crypto (Zkn{ed}) is not supported. Most of the rest of the `soft`
+// is unused which generates warnings about unused definitions. We just disable
+// the warnings here rather than cluttering `soft` with RISC-V conditionals.
+allow(unused))]
 mod soft;
 
 use cfg_if::cfg_if;
@@ -138,6 +159,9 @@ cfg_if! {
         mod armv8;
         mod autodetect;
         pub use autodetect::*;
+    } else if #[cfg(all(target_arch = "riscv64", target_feature = "zkne", target_feature = "zknd"))] {
+        mod riscv;
+        pub use riscv::rv64::*;
     } else if #[cfg(all(
         any(target_arch = "x86", target_arch = "x86_64"),
         not(aes_force_soft)
--- a/aes/src/riscv.rs
+++ b/aes/src/riscv.rs
@@ -1,0 +1,149 @@
+//! AES block cipher implementations for RISC-V using the Cryptography
+//! Extensions
+//!
+//! Supported targets: rv64 (scalar)
+//!
+//! NOTE: rv32 (scalar) is not currently implemented, primarily due to the
+//! difficulty in obtaining a suitable development environment (lack of distro
+//! support and lack of precompiled toolchains), the effort required for
+//! maintaining a test environment as 32-bit becomes less supported, and the
+//! overall scarcity of relevant hardware. If someone has a specific need for
+//! such an implementation, please open an issue.
+//!
+//! NOTE: These implementations are currently not enabled through
+//! auto-detection. In order to use this implementation, you must enable the
+//! appropriate target-features.
+//!
+//! Examining the module structure for this implementation should give you an
+//! idea of how to specify these features in your own code.
+//!
+//! NOTE: AES-128, AES-192, and AES-256 are supported.
+
+#[cfg(all(
+    target_arch = "riscv64",
+    target_feature = "zknd",
+    target_feature = "zkne"
+))]
+pub(crate) mod rv64;
+
+use crate::Block;
+
+#[cfg(test)]
+mod test {
+    use hex_literal::hex;
+
+    #[cfg(all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_feature = "zknd",
+        target_feature = "zkne"
+    ))]
+    mod inner {
+        use super::*;
+
+        pub(crate) const AES128_EXP_INVKEYS: [[u8; 16]; 11] = [
+            AES128_KEY,
+            hex!("2b3708a7f262d405bc3ebdbf4b617d62"),
+            hex!("cc7505eb3e17d1ee82296c51c9481133"),
+            hex!("7c1f13f74208c219c021ae480969bf7b"),
+            hex!("90884413d280860a12a128421bc89739"),
+            hex!("6ea30afcbc238cf6ae82a4b4b54a338d"),
+            hex!("6efcd876d2df54807c5df034c917c3b9"),
+            hex!("12c07647c01f22c7bc42d2f37555114a"),
+            hex!("df7d925a1f62b09da320626ed6757324"),
+            hex!("0c7b5a631319eafeb0398890664cfbb4"),
+            hex!("d014f9a8c9ee2589e13f0cc8b6630ca6"),
+        ];
+
+        pub(crate) const AES192_KEY: [u8; 24] =
+            hex!("8e73b0f7da0e6452c810f32b809079e562f8ead2522c6b7b");
+        pub(crate) const AES192_EXP_KEYS: [[u8; 16]; 13] = [
+            hex!("8e73b0f7da0e6452c810f32b809079e5"),
+            hex!("62f8ead2522c6b7bfe0c91f72402f5a5"),
+            hex!("ec12068e6c827f6b0e7a95b95c56fec2"),
+            hex!("4db7b4bd69b5411885a74796e92538fd"),
+            hex!("e75fad44bb095386485af05721efb14f"),
+            hex!("a448f6d94d6dce24aa326360113b30e6"),
+            hex!("a25e7ed583b1cf9a27f939436a94f767"),
+            hex!("c0a69407d19da4e1ec1786eb6fa64971"),
+            hex!("485f703222cb8755e26d135233f0b7b3"),
+            hex!("40beeb282f18a2596747d26b458c553e"),
+            hex!("a7e1466c9411f1df821f750aad07d753"),
+            hex!("ca4005388fcc5006282d166abc3ce7b5"),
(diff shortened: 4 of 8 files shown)


aes/src/lib.rs
```diff
@@ -123,12 +130,26 @@
...
+)]
 
 #[cfg(feature = "hazmat")]
 pub mod hazmat;
 
 #[macro_use]
 mod macros;
+#[cfg_attr(any(target_arch = "riscv32", target_arch = "riscv64"),
+// The `soft` AES-192 is used for RISC-V when RVV crypto (Zvkned) is supported
+// but Scalar crypto (Zkn{ed}) is not supported. Most of the rest of the `soft`
```
newpavlov, line 149: This PR is for scalar crypto support, so it looks like this part is not needed.

aes/src/lib.rs
```diff
@@ -123,12 +130,26 @@
 )]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![warn(missing_docs, rust_2018_idioms)]
+#![cfg_attr(
+    all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_feature = "zknd",
+        target_feature = "zkne"
+    ),
+    feature(riscv_ext_intrinsics)
```
newpavlov, line 139: Please use the same approach as in the `sha2` crate, i.e. introduce a configuration flag which unconditionally enables the scalar backend, relies on Nightly, and requires user to enable the target features. It's not correct to switch to Nightly-only based on enabled target features.

aes/benches/mod.rs
```diff
@@ -13,11 +13,19 @@ block_decryptor_bench!(
     aes128_decrypt_block,
     aes128_decrypt_blocks,
 );
+#[cfg(any(
+    not(target_arch = "riscv64"),
+    all(target_feature = "zknd", target_feature = "zkne")
+))]
```
newpavlov, line 19: Remove these `cfg`s.

.github/workflows/aes.yml
```diff
@@ -310,6 +310,33 @@ jobs:
...
+    defaults:
+      run:
+        working-directory: .
+    steps:
+      - uses: actions/checkout@v4
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
+      - run: cargo install cross --git https://github.com/cross-rs/cross
```
newpavlov, line 333: Use our CI job for cross installation which install a pre-compiled version of `cross`.

aes/src/riscv.rs
```diff
@@ -0,0 +1,149 @@
...
+#[cfg(test)]
+mod test {
+    use hex_literal::hex;
+
+    #[cfg(all(
+        any(target_arch = "riscv32", target_arch = "riscv64"),
+        target_feature = "zknd",
+        target_feature = "zkne"
+    ))]
+    mod inner {
```
newpavlov, line 40: Why do you need this module?

newpavlov (changes requested):
I don't quite like the implementation in it's current form. I will need to read the code a bit more thoroughly to give concrete suggestions. You may have wrote it while keeping the RVV support in mind, but I think we should make a clear scalar implementation first. Potential generalization should be done in later PRs.

aes/src/lib.rs
silvanshade: This is still needed to silence warnings about unused definitions for architectures without `autodetect` support (because `soft` will be unused in that case) but the reasoning behind the comment is confusing now that the vector support has been separated out.

I've changed the conditional to this:

```rust
#[cfg_attr(not(any(
    target_arch = "aarch64",
    target_arch = "x86",
    target_arch = "x86_64",
)),
// Architectures that have hardware-support backends, but that do not use
// `autodetect` -- like RISC-V 64 -- will generate warnings about unused
// definitions in `soft`. We silence them once here in order to avoid cluttering
// `soft` with architecture specific conditionals.
allow(unused))]
mod soft;
```

aes/src/riscv.rs
silvanshade: This was a relic from earlier refactoring where I was trying to silence warnings about unused definitions when the scalar AES-192 key expansion was being used. That required a more granular approach with nested modules like this. But I didn't like having to factor it this way which is why I ultimate switched to a module-level `#[allow(unused)]` even though it's less precise. I just forgot to remove this one inner module.

silvanshade:
> You may have wrote it while keeping the RVV support in mind, but I think we should make a clear scalar implementation first. Potential generalization should be done in later PRs.

Aside from vestiges in comments and conditionals referring to vector support which you pointed out (and I've since removed), there isn't anything intentional in the design of the scalar implementation which is influenced by the vector implementation.

I'm not sure what you're referring to regarding generalizations.

The scalar implementation here is "idiomatic", for lack of a better description, and appears less straightforward than what one might expect.

That isn't because I wanted to write it this way. It is a reflection of the nature of RISC-V scalar crypto instructions:

<details>
aes
</details>

Due to those semantics, where the state of the cipher is staggered by half-rounds in this way, you end up with code that looks like this:

```rust
#[inline(always)]
pub(super) fn enc1_two_more(&mut self, k0: RoundKey, k1: RoundKey) {
    let mut n0;
    let mut n1;
    self.data[0] ^= k0[0];
    self.data[1] ^= k0[1];
    n0 = unsafe { aes64esm(self.data[0], self.data[1]) };
    n1 = unsafe { aes64esm(self.data[1], self.data[0]) };
    n0 ^= k1[0];
    n1 ^= k1[1];
    self.data[0] = unsafe { aes64esm(n0, n1) };
    self.data[1] = unsafe { aes64esm(n1, n0) };
}
```

I don't think that code is very easy to understand at a glance without already being familiar with the semantics of the instructions.

That's why I tried to abstract it away into a clearer, although less direct interface:

```rust
pub(super) struct CipherState1 {
    data: [u64; 2],
}

impl CipherState1 {
    #[inline(always)]
    pub(super) fn load1(block: &Block) -> Self {
        let ptr = block.as_ptr().cast::<u64>();
        let s0 = unsafe { ptr.add(0).read_unaligned() };
        let s1 = unsafe { ptr.add(1).read_unaligned() };
        Self { data: [s0, s1] }
    }

    #[inline(always)]
    pub(super) fn save1(self, block: &mut Block) {
        let b0 = self.data[0].to_ne_bytes();
        let b1 = self.data[1].to_ne_bytes();
        block[00..08].copy_from_slice(&b0);
        block[08..16].copy_from_slice(&b1);
    }

    #[inline(always)]
    pub(super) fn xor1(&mut self, key: &RoundKey) {
        self.data[0] ^= key[0];
        self.data[1] ^= key[1];
    }

    #[inline(always)]
    pub(super) fn enc1_two_more(&mut self, k0: RoundKey, k1: RoundKey) {
        let mut n0;
        let mut n1;
        self.data[0] ^= k0[0];
        self.data[1] ^= k0[1];
        n0 = unsafe { aes64esm(self.data[0], self.data[1]) };
        n1 = unsafe { aes64esm(self.data[1], self.data[0]) };
        n0 ^= k1[0];
        n1 ^= k1[1];
        self.data[0] = unsafe { aes64esm(n0, n1) };
        self.data[1] = unsafe { aes64esm(n1, n0) };
    }

    #[inline(always)]
    pub(super) fn enc1_two_last(&mut self, k0: RoundKey, k1: RoundKey) {
        let mut n0;
        let mut n1;
        self.data[0] ^= k0[0];
        self.data[1] ^= k0[1];
        n0 = unsafe { aes64esm(self.data[0], self.data[1]) };
        n1 = unsafe { aes64esm(self.data[1], self.data[0]) };
        n0 ^= k1[0];
        n1 ^= k1[1];
        self.data[0] = unsafe { aes64es(n0, n1) };
        self.data[1] = unsafe { aes64es(n1, n0) };
    }

    #[inline(always)]
    pub(super) fn dec1_two_more(&mut self, k0: RoundKey, k1: RoundKey) {
        let mut n0;
        let mut n1;
        n0 = unsafe { aes64dsm(self.data[0], self.data[1]) };
        n1 = unsafe { aes64dsm(self.data[1], self.data[0]) };
        self.data[0] = n0 ^ k1[0];
        self.data[1] = n1 ^ k1[1];
        n0 = unsafe { aes64dsm(self.data[0], self.data[1]) };
        n1 = unsafe { aes64dsm(self.data[1], self.data[0]) };
        self.data[0] = n0 ^ k0[0];
        self.data[1] = n1 ^ k0[1];
    }

    #[inline(always)]
    pub(super) fn dec1_two_last(&mut self, k0: RoundKey, k1: RoundKey) {
        let mut n0;
        let mut n1;
        n0 = unsafe { aes64dsm(self.data[0], self.data[1]) };
        n1 = unsafe { aes64dsm(self.data[1], self.data[0]) };
        self.data[0] = n0 ^ k1[0];
        self.data[1] = n1 ^ k1[1];
        n0 = unsafe { aes64ds(self.data[0], self.data[1]) };
        n1 = unsafe { aes64ds(self.data[1], self.data[0]) };
        self.data[0] = n0 ^ k0[0];
        self.data[1] = n1 ^ k0[1];
    }
}
```

The situation is similar for the key-expansion instructions.

If this is what you were referring to regarding generalization, we could remove the abstraction and make the code more direct, but I think that would make it harder to understand and harder to maintain.

aes/src/riscv/rv64/encdec.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        #[inline(always)]
+        pub(super) fn enc1_two_more(&mut self, k0: RoundKey, k1: RoundKey) {
+            let mut n0;
+            let mut n1;
+            self.data[0] ^= k0[0];
+            self.data[1] ^= k0[1];
+            n0 = unsafe { aes64esm(self.data[0], self.data[1]) };
+            n1 = unsafe { aes64esm(self.data[1], self.data[0]) };
```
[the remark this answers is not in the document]
silvanshade: You're right that it's generally better to use `target_feature`.

The reason I did it this way is because `#[target_feature(enable = "zkne")]` is incompatible with `#[inline(always)]` and we already gate the outer module on the target feature.

But I guess even with `inline(always)` we can't actually guarantee it gets inlined so I'll change it.

aes/src/riscv/rv64/expand.rs
```diff
@@ -0,0 +1,217 @@
+#![allow(clippy::identity_op)]
+
+use crate::riscv::rv64::{RoundKey, RoundKeys};
+use core::{arch::riscv64::*, mem::MaybeUninit};
+
+// TODO(silvanshade): `COLUMNS` should be an associated constant once support for that is stable.
+pub(crate) struct KeySchedule<const COLUMNS: usize, const ROUNDS: usize> {
+    cols: [u64; COLUMNS],
+    keys: [MaybeUninit<RoundKey>; ROUNDS],
```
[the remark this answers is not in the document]
silvanshade: I rewrote the code to eliminate the abstraction and remove the use of `MaybeUninit`.

This is how it looks now (below). I don't really think this is clearer but it is more direct I guess.

It has to be unrolled like this and can't be written in a loop because:

1. `aes64ks1i` takes a constant parameter `RNUM`.
2. `aes64ks1i` is not itself a `const fn`, so cannot be used in a `const` loop.

```rust
    #[inline]
    #[target_feature(enable = "zknd", enable = "zkne")]
    pub(crate) fn expand_key(key: &[u8; 16]) -> RoundKeys<11> {
        let (word_bytes, tail) = key.as_chunks::<8>();
        assert!(tail.is_empty());

        let mut cols = core::array::from_fn::<u64, 2, _>(|i| u64::from_ne_bytes(word_bytes[i]));
        let mut keys = <[RoundKey; 11]>::default();

        keys[0][0] = cols[0];
        keys[0][1] = cols[1];

        let s = aes64ks1i(cols[1], 0);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[1][0] = cols[0];
        keys[1][1] = cols[1];

        let s = aes64ks1i(cols[1], 1);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[2][0] = cols[0];
        keys[2][1] = cols[1];

        let s = aes64ks1i(cols[1], 2);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[3][0] = cols[0];
        keys[3][1] = cols[1];

        let s = aes64ks1i(cols[1], 3);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[4][0] = cols[0];
        keys[4][1] = cols[1];

        let s = aes64ks1i(cols[1], 4);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[5][0] = cols[0];
        keys[5][1] = cols[1];

        let s = aes64ks1i(cols[1], 5);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[6][0] = cols[0];
        keys[6][1] = cols[1];

        let s = aes64ks1i(cols[1], 6);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[7][0] = cols[0];
        keys[7][1] = cols[1];

        let s = aes64ks1i(cols[1], 7);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[8][0] = cols[0];
        keys[8][1] = cols[1];

        let s = aes64ks1i(cols[1], 8);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[9][0] = cols[0];
        keys[9][1] = cols[1];

        let s = aes64ks1i(cols[1], 9);
        cols[0] = aes64ks2(s, cols[0]);
        cols[1] = aes64ks2(cols[0], cols[1]);

        keys[10][0] = cols[0];
        keys[10][1] = cols[1];

        keys
    }
```

aes/src/riscv/rv64/expand.rs
```diff
@@ -0,0 +1,293 @@
+#![allow(clippy::identity_op)]
+
+use crate::riscv::rv64::{RoundKey, RoundKeys};
+use core::arch::riscv64::*;
+
+pub(crate) struct KeySchedule<const COLUMNS: usize, const ROUNDS: usize>;
```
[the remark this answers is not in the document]
silvanshade: The struct provides a dispatch mechanism through the const parameters which is a cleaner approach than adding identifiers as arguments to the macro.

aes/src/riscv/rv64/expand.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        keys[0][0] = cols[0];
+        keys[0][1] = cols[1];
+
+        let s = aes64ks1i(cols[1], 0);
+        cols[0] = aes64ks2(s, cols[0]);
+        cols[1] = aes64ks2(cols[0], cols[1]);
+
+        keys[1][0] = cols[0];
+        keys[1][1] = cols[1];
```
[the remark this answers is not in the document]
silvanshade: I had several functions like that previously and you said it was over-engineered so I removed them.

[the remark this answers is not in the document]
newpavlov: The over-engineering remark was primarily about the struct and use of `MaybeUninit`.

aes/src/riscv/rv64/expand.rs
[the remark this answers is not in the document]
newpavlov: IMO passing an identifier is a more straightforward approach, but ok we can leave it as-is for now. I probably will experiment with this part of the code later in a separate PR.

aes/src/riscv/rv64/expand.rs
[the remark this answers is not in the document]
newpavlov: Ok.

newpavlov:
Regarding optimal number of parallel blocks. I played a bit with [this snippet](https://rust.godbolt.org/z/Gn7hM8rGh) which implements simple ECB encryption on aligned buffers. I measured a number of operations which perform stack-based store/load inside loop and got the following results:
| Blocks | Aes-128 | Aes-192 | Aes-256
|:--:|:--:|:--:|:--:|
| 1 | 0 | 6 | 10 |
| 2 | 4 | 10 | 11 |
| 3 | 5 | 12 | 13 |
| 4 | 9 | 14 | 16 |
| 5 | 10 | 14 | 18 |
| 6 | 13 | 18 | 22 |
| 7 | 15 | 19 | 26* |
| 8 | 20 | 47* | 59* |
| 9 | 62 | 70* | 90* |

Note that the results marked with asterisk required manual unrolling of the loop. It looks like 8 blocks is a good number for AES-128, but for AES-192 and AES-256 it may be worth to lower it. I think the compiler could do a slightly better job with its register allocations, but alas.

On RV32 number of parallel blocks should be probably 4 or smaller.

In practice, the results will be much worse because of the horrible handling of misaligned loads/stores.

aes/src/riscv/rv64/encdec.rs
```diff
@@ -0,0 +1,72 @@
...
+        utils::{AlignedBlock, AlignedParBlock},
+    },
+};
+use cipher::inout::InOut;
+
+#[inline]
+#[target_feature(enable = "zkne")]
+pub(super) fn encrypt1<const N: usize>(keys: &RoundKeys<N>, mut block1: InOut<'_, '_, Block>) {
+    let mut block = AlignedBlock::load(block1.get_in());
+    #[allow(clippy::incompatible_msrv)]
```
newpavlov, line 14: Move these `allow`s to the `riscv` or `rv64` module.

aes/src/riscv/rv64/expand.rs
```diff
@@ -0,0 +1,283 @@
+#![allow(clippy::identity_op)]
```
newpavlov, line 1: Is this `allow` still needed?

newpavlov:
Also note that in the godbolt snippet above the compiler annoyingly and absolutely unnecessarily spills round keys to stack. It's a [known problem](https://github.com/llvm/llvm-project/issues/53348), but there is zero work on it in the years since it was reported, so I wouldn't bet on it being resolved anytime soon...

PR #523, opened 2026-01-20
feat: Implement Simon

sorairolake:
This pull request adds an implementation of the [Simon block cipher](https://en.wikipedia.org/wiki/Simon_(cipher)).

I referred to <https://eprint.iacr.org/2013/404.pdf> for the algorithm and test vectors. The source code is derived from the [`speck-cipher`](https://crates.io/crates/speck-cipher) crate, which implements the sister algorithm, Speck.

Partially solves #1

The name of this crate is [`simon-cipher`](https://crates.io/crates/simon-cipher) because it is already registered by @newpavlov. the [`simon`](https://crates.io/crates/simon) crate already exists for other purposes.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,7 @@ on:
       'rc5-v*',
       'rc6-v*',
       'serpent-v*',
+      'simon-cipher-v*',
       'sm4-v*',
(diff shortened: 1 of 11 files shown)


sorairolake:
@newpavlov @tarcieri Could you please review this pull request?

Issue #559, opened 2026-05-13
aes: SIMD-acceleration for WASM


sinui0:
WASM targets currently use the software fixslice implementation which processes up to 4 blocks at a time (fixslice64). WebAssembly has had the `simd128` proposal stabilized in the spec for a while now, and the corresponding intrinsics ([`core::arch::wasm32::*`]) have been stable in Rust since 1.33. Every major WebAssembly runtime ships `simd128` support today.

We use AES extensively in our projects which target running in browsers, the throughput of AES in WASM is a major performance lever for us. Translating the existing fixslice64 implementation (AI assisted, not sure what your policy is on that) to use 128 bit lanes yields the expected >2x throughput increase in V8 (chromium), and >3x in SpiderMonkey (firefox).

(AMD Ryzen 7 5800X)

|Unit: MB/s | wasmtime soft | wasmtime simd128 | **wasmtime relative** | v8 soft | v8 simd128 | **v8 relative** | sm soft | sm simd128 | **sm relative** |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| AES-128 enc | 186 | 184 | 0.99× | 258 | **604** | **2.34×** | 213 | **657** | **3.08×** |
| AES-128 dec | 185 | 122 | **0.66×** | 235 | 508 | 2.16× | 182 | **598** | **3.29×** |
| AES-192 enc | 159 | 150 | 0.94× | 224 | 522 | 2.33× | 185 | **588** | **3.18×** |
| AES-192 dec | 159 | 100 | **0.63×** | 202 | 486 | 2.41× | 153 | **577** | **3.77×** |
| AES-256 enc | 138 | 140 | 1.01× | 192 | 455 | 2.37× | 159 | **491** | **3.09×** |
| AES-256 dec | 135 | 89 | **0.66×** | 171 | 424 | 2.48× | 134 | **453** | **3.38×** |

As you can see, this causes a regression in Wasmtime decrypt performance due to bad JIT by Cranelift. I haven't been able to mitigate that without incurring significant throughput loss in the browser runtimes. I'm not sure whether the right approach is to eat the regression by default, or to put the SIMD behind a flag.

[Implementation](https://github.com/privacy-ethereum/block-ciphers/commit/c2d51cbaf03ab1c51b2ac221153eb893f6a1b891)

Can reproduce the benchmarks using [wasm-harness](https://crates.io/crates/wasm-harness)

[...]

PR #560, opened 2026-05-15
refactor(aes): machine-word generic fixslice

sinui0:
This PR collapses the two fixslice impls into a single `fixslice.rs` parameterized over a `Word` trait that abstracts the machine-word width.

I tried to structure the commits to make it clear to the reviewer that this introduces no changes to the existing implementation logic.

This PR was motivated by #559 to support larger machine-word (SIMD) implementations without introducing major duplication and maintenance burden.


tarcieri:
This looks good to me as a general direction, though it may be some time before I can review it in detail.

I'm a bit confused how this broke the build the way it did, it seems somehow with only changes to the soft backend it broke things in the NEON backend somehow?

sinui0:
I haven't looked closely at the CI yet, but perhaps it is already broken on latest stable rustc prior to this PR?

newpavlov:
The CI failures were fixed in #562.

aes/src/soft/fixslice.rs
```diff
@@ -16,30 +22,135 @@
...
+    + BitXorAssign
+    + Not<Output = Self>
+    + Shl<u32, Output = Self>
+    + Shr<u32, Output = Self>
+{
+    /// Number of 128-bit blocks bitsliced together in one state.
+    type Blocks: ArraySize;
+
+    /// Width in bits of one row of the bitsliced state (8 for `u32`, 16 for `u64`).
+    const ROW_BITS: u32;
```
tarcieri, line 50: Seems like this could have a default? (could probably use an explanatory comment)

```suggestion
    const ROW_BITS: u32 = size_of::<Self>() * 2;
```

...or if `HALF_ROW` were defined this way it gets rid of the `* 2`

tarcieri (approved):
I really like the amount of duplication this gets rid of and would like to see it landed

aes/src/soft/fixslice.rs
newpavlov: Huh, I am a bit surprised that it works. I will fix it as part of #575.

PR #562, opened 2026-05-25
aes: fix min version of `zeroize` and build warnings on AArch64

newpavlov:


aes/src/armv8/encdec.rs
```diff
@@ -68,7 +68,7 @@ pub(super) unsafe fn encrypt_par<const KEYS: usize, ParBlocks: ArraySize>(
     keys: &[uint8x16_t; KEYS],
     blocks: InOut<'_, '_, Array<Block, ParBlocks>>,
 ) {
-    #[inline(always)]
+    #[target_feature(enable = "aes")]
```
newpavlov, line 71: The function gets properly inlined without `#[inline(always)]`, see: https://rust.godbolt.org/z/de18q6dav

PR #563, opened 2026-05-25
Release aes v0.9.1

newpavlov:
### Fixed
- Minimal version of `zeroize` dependency ([#562])
- Build warnings on AArch64 targets ([#562])

[#562]: https://github.com/RustCrypto/block-ciphers/pull/562

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "aes"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "cipher",
  "cpubits",
--- a/aes/CHANGELOG.md
+++ b/aes/CHANGELOG.md
@@ -5,14 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.9.1 (UNRELEASED)
+## 0.9.1 (2026-05-26)
 ### Fixed
 - Minimal version of `zeroize` dependency ([#562])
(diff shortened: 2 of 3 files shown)


newpavlov:
@tarcieri 
Note that I think it's worth to yank v0.9.0 to fix minimal versions build in downstream crates relying on the `zeroize` feature.

tarcieri (approved):

dvzrv:
> Note that I think it's worth to yank v0.9.0 to fix minimal versions build in downstream crates relying on the `zeroize` feature.

Hm, could you elaborate why yanking the 0.9.0 version was necessary? It is quite disruptive to downstream CI as well :sweat_smile:

newpavlov:
Because we test minimal dependency versions as part of our CI tests, so downstream crates could pull `aes` v0.9.0 and incorrect version of `zeroize` which results in a compilation error since `aes` uses function introduced only in `zeroize` v1.8. So we either have to pin aes to v0.9.1 in _all_ downstream crates or simply yank v0.9.0 as a "buggy" version.

>It is quite disruptive to downstream CI as well

It should not be disruptive. If you commit Cargo.lock, then yanking should not affect you, if not, then v0.9.1 should be pulled by default.

joshsleeper:
> It should not be disruptive. If you commit Cargo.lock, then yanking should not affect you,

huh, I was gonna say that it would _specifically and explicitly_ affect folks who commit and install from `Cargo.lock` if said lockfile has `v0.9.0` locked... but TIL `cargo` tolerates locked + yanked packages with a warning to avoid breakage apparently

> Cargo will not use a yanked version for any new project or checkout without a pre-existing lockfile, and will generate an error if there are no longer any compatible versions for your crate.

<https://doc.rust-lang.org/cargo/commands/cargo-yank.html>

PR #569, opened 2026-06-22
aes: use `intel-sde-install` action for AVX-512 CI

tarcieri:


tarcieri:
Cool, looks like that did the trick

PR #575, opened 2026-07-21
aes: refactor backends

### LINKED ISSUE #574: aes: Performance regressions from aes-0.8.x to 0.9.1 presumably due to differences in inlining on x86_64
We have recently started to migrate Fuchsia to aes 0.9.1 and see some performance regression with it compared to aes 0.8.4 on x86_64.

Most regressions we observe are in the range of 3% to 8%, although a few outliers are significantly larger than that (two-digit percentages). Disassembly shows that there are differences in the inlining between the different versions.

It seems there is no open issue referencing this problem, but it is unlikely that this is a Fuchsia-specific problem.

newpavlov:
Simplifies and consolidates switching between backends. Fixes performance regression introduced in v0.9. 

Closes #574


aes/src/lib.rs
```diff
@@ -125,32 +125,417 @@
...
+        }
+    };
+}
+
+macro_rules! common_impls {
+    ($name:ident) => {
+        impl Clone for $name {
+            #[inline]
+            fn clone(&self) -> Self {
+                unsafe { core::ptr::read(self) }
```
tarcieri: This feels a little brittle. I'm surprised it isn't triggering a lint.

aes/src/backends.rs
```diff
@@ -0,0 +1,24 @@
...
+pub(crate) mod aarch64_aes;
+
+#[cfg(all(
+    any(target_arch = "x86_64", target_arch = "x86"),
+    not(aes_backend = "soft")
+))]
+pub(crate) mod x86_aes;
+
+#[cfg(all(
+    any(aes_backend = "avx256", aes_backend = "avx256"),
```
tarcieri: This seems to be duplicated?
```suggestion
    aes_backend = "avx256",
```

tarcieri (approved):
This looks like a nice cleanup that eliminates a lot of duplication.

My personal preference would be to land #560 first, especially as it was opened first. The two PRs together would eliminate quite a bit of duplication.

aes/src/backends.rs
newpavlov: It should've been `any(aes_backend = "avx256", aes_backend = "avx512")`.

aes/src/lib.rs
newpavlov: AFAIK it should work without issues for Plain Old Data. Considering that the backend states implement `Copy`, I don't think it will be easy to accidentally screw this part.

Without relying on `ptr::read` we would need to use `cfg_if` here as well and I tried to minimize their number as much as possible.

newpavlov: Switched to `#[derive(Clone)]` since it's possible to derive `Copy` for `union`.

Issue #578, opened 2026-07-24
aes: reduce state size of software backends


newpavlov:
Currently the fixslice backend keeps "broadcasted" round keys as part of its state. For example, on 64-bit targets it results in a 704 byte state for AES-128, which is 4 times bigger relative to the other backends. With hypothetical future SIMD backends (see #559) it would be even worse. It also results in a bloated block cipher state even if the software backend is not used (e.g. when project is compiled with `target-feature=+aes`).

So it's probably worth to use a more compact representation of round keys and broadcast them only during backend creation like we do with VAES backends.

tarcieri:
FWIW there’s #191. Definitely discussed this with @briansmith in the past

newpavlov:
Ah, true. I will close it as a duplicate then.

Issue #191, opened 2020-10-29
Optional compact round keys for AES fixslice


peterdettman:
In the current fixslice implementation there is some redundancy in the way keys are stored (2x for fixslice32, 4x for fixslice64). This enables the round keys to simply be XORed in for each round.

It is possible to store the round keys in a compact format and expand them in each round, which requires a few additional instructions (prototype code for fixslice64 puts this at 10% slower; the penalty will be smaller for fixslice32). The memory saving for the key storage is then: 50% i.e. 176/208/240 bytes for fixslice32 AES 128/192/256 respectively, or 75% i.e. 528/624/720 bytes for fixslice64.

Is there a standard cfg attribute for constrained memory devices, and/or should this project support a feature making smaller memory footprint a priority?

newpavlov:
In some crates we have `no_unroll` feature (e.g. [`kuznyechik`](https://github.com/RustCrypto/block-ciphers/blob/master/kuznyechik/Cargo.toml#L25) and [`keccak`](https://github.com/RustCrypto/sponges/blob/master/keccak/Cargo.toml)), which trades performance for reduced binary size.

It looks like expanding round keys per each `encrypt_blocks` call will be quite inefficient. We probably should first introduce a method which will process `&mut [Block<Self>]`.

tarcieri:
As of the latest release of `cipher` (v0.2.5) there are now `BlockCipher::{encrypt_slice, decrypt_slice}` methods:

https://docs.rs/cipher/0.2.5/cipher/block/trait.BlockCipher.html#method.encrypt_slice

These would allow you to expand the round keys before operating over a sequence of blocks.

briansmith:
@peterdettman wrote:
> It is possible to store the round keys in a compact format and expand them in each round, which requires a few additional instructions (prototype code for fixslice64 puts this at 10% slower; the penalty will be smaller for fixslice32). The memory saving for the key storage is then: 50% i.e. 176/208/240 bytes for fixslice32 AES 128/192/256 respectively, or 75% i.e. 528/624/720 bytes for fixslice64.

Put another way, implementing this suggestion would make the expanded keys the same size they are for AES-NI.

[...]

PR #593, opened 2026-09-08
aes: disable soft backend if not needed

Freax13:
Previously, the soft backend was always enabled even when the target supports one of the accelerated backends. This is a problem because the soft backend uses a lot of memory for the round keys.

Disabling the soft backend when it's not needed saves a lot of memory: On a modern AMD64 system, Aes256Enc now only requires 240 bytes instead of 960 bytes.


newpavlov:
I would strongly prefer to implement #191 instead of doing this (also see #578). I especially do not want to introduce build scripts to `aes`.

Freax13:
> I would strongly prefer to implement #191 instead of doing this (also see #578).

Fair enough.

> I especially do not want to introduce build scripts to `aes`.

Initially I implemented this without a build script, but found this to be a bit too verbose: https://github.com/RustCrypto/block-ciphers/commit/d2143b31e26cb7681ca9e3e276b9ba7b6f008bcd.

Let me know if you prefer that approach.

newpavlov:
>Initially I implemented this without a build script, but found this to be a bit too verbose

Yes, this is why we did not implement it as well despite being aware of the problem. :)

PR #594, opened 2026-09-10
aes: shrink AES round keys for AES fixslice backend

### LINKED ISSUE #191: Optional compact round keys for AES fixslice

Freax13:
This PR does two things:
1. Add support for using u16 as the native word size for the AES fixslice backend.
2. Round keys are now always store as `u16`s. They are expanded into the full native words when the backend is created.
  a. On a targets with a 64-bit native word size, this reduces the size of the `Aes128`, `Aes192`, and `Aes256` types from 704, 832, and 960 to 176, 208, and 240 bytes respectively.

Closes #191


aes/src/backends/fixslice/mod.rs
```diff
@@ -31,7 +31,10 @@ use word::Word;
 type State<W> = [W; 8];
 
 cpubits::cpubits! {
-    16 | 32 => {
+    16 => {
+        pub(super) type NativeWord = u16;
+    }
```
tarcieri, line 36: Have you actually tested this, and if so on what?

It'd be nice if we are trying to actually support 16-bit targets to at least ensure code compiles on them in CI

newpavlov:
I don't think we need to introduce `u16`-based backend for compact keys. Instead it would be better to store compact round keys in the `Aes` state and perform their expansion during backend creation (see the AVX-512 backend for reference).

aes/src/backends/fixslice/mod.rs
Freax13: > Have you actually tested this, and if so on what?

I did not test this on a 16-bit target, but I did test this by manually changing `NativeWord` to `u16` in the `64 => {...}` branch.

> It'd be nice if we are trying to actually support 16-bit targets to at least ensure code compiles on them in CI

I suppose we could run miri with a 16-bit target. That should exercise the 16-bit code.

Freax13:
> I don't think we need to introduce `u16`-based backend for compact keys.

Just to be clear, even with the compact backend, the fixslice implementation still runs on the native word size. Enabling the compact backend does not force it into the u16 path for encryption/decryption.

> Instead it would be better to store compact round keys in the `Aes` state and perform their expansion during backend creation (see the AVX-512 backend for reference).

The proposed changes are very similar to this except that we expand/broadcast the round keys one by one when they're needed instead of doing so at the start of the `encrypt`/`decrypt` function. Expanding them at the beginning of `encrypt`/`decrypt` would be possible, but it would require saving all the expanded state on the stack and negate the space savings that motivated this change in the first place.

newpavlov:
>it would require saving all the expanded state on the stack and negate the space savings that motivated this change in the first place.

Yes, broadcasted keys are likely to be spilled to the stack (at least on register-starved x86), but it would reduce size of the `Aes` state (so we would not need bother with something like #593) and would have much less runtime overhead since broadcast has to be done only once during backend selection.

aes/src/backends/fixslice/mod.rs
Freax13: It turns out there's only a single 16-bit target, msp430-none-elf, and it's a no_std target, so it can't run tests.

I added a matrix job with `--cfg cpubits="16"` to test 16-bit support.

Freax13:
> > it would require saving all the expanded state on the stack and negate the space savings that motivated this change in the first place.
> 
> Yes, broadcasted keys are likely to be spilled to the stack (at least on register-starved x86), but it would reduce size of the `Aes` state (so we would not need bother with something like #593) and would have much less runtime overhead since broadcast has to be done only once during backend selection.

Done.

.github/workflows/aes.yml
```diff
@@ -203,24 +203,36 @@ jobs:
...
             rust: 1.89.0 # MSRV
+            RUSTFLAGS: '-Dwarnings --cfg aes_backend="soft"'
+          - target: x86_64-unknown-linux-gnu
+            rust: stable
+            RUSTFLAGS: '-Dwarnings --cfg aes_backend="soft"'
+
+          # 64-bit Linux with cpubits override
+          - target: x86_64-unknown-linux-gnu
+            rust: 1.89.0 # MSRV
+            RUSTFLAGS: '-Dwarnings --cfg aes_backend="soft" --cfg cpubits="16"'
```
newpavlov, line 230: I think it's better to introduce a separate job for testing the 16 bit backend instead of modifying the `soft` job until we will be able to test the backend on a true 16 bit target.

aes/src/backends/fixslice/mod.rs
```diff
@@ -41,3 +44,4 @@ cpubits::cpubits! {
 
 pub(super) type NativeBatchSize = <NativeWord as Word>::Blocks;
 pub(super) type BatchBlocks<W> = Array<crate::Block, <W as Word>::Blocks>;
+pub(super) type RoundKeyWord = u16;
```
newpavlov: Maybe name it as something like `MinWord`?

aes/src/backends/soft.rs
```diff
@@ -24,53 +28,79 @@ macro_rules! impl_backend {
...
-                let res = fixslice::$module::decrypt(&self.keys, &blocks);
+                let res = fixslice::$module::decrypt(&self.rk, &blocks);
                 *block.get_out() = res[0];
             }
 
             #[inline(always)]
             fn decrypt_par_blocks(&self, mut blocks: InOut<'_, '_, ParBlocks<Self>>) {
-                let res = fixslice::$module::decrypt(&self.keys, blocks.get_in());
+                let res =
+                    fixslice::$module::decrypt::<NativeWord>(&self.rk_native, blocks.get_in());
```
newpavlov: Do we need the turbofish here? I think the compiler should be able to infer the type from `rk_native`.

aes/src/backends/soft.rs
```diff
@@ -24,53 +28,79 @@ macro_rules! impl_backend {
...
+                f.call(&backend)
             }
         }
 
-        impl BlockSizeUser for $name {
+        #[doc=$doc]
+        #[doc = "block cipher"]
+        #[derive(Clone, Copy)]
+        pub(crate) struct $backend<'a> {
+            rk: &'a fixslice::$module::RoundKeys<u16>,
```
newpavlov: Use `RoundKeyWord` here.

newpavlov (commented):
I haven't checked the changes in fixslice internals, but it looks mostly good otherwise.

Ideally, it would be nice to keep `aes192`, `mix_columns`, and `utils` intact, but I am fine with the changes if there is a good reason for them.

.github/workflows/aes.yml
Freax13: The git diff is a bit confusing for this. I did only add new items to the matrix we use to test the soft backend. I did not modify existing items in the matrix. Or did you want me to add an entirely new job for this?

newpavlov: Yes, I meant an entirely new job just for testing 16 bit.

aes/src/backends/fixslice/mod.rs
Freax13: Done

aes/src/backends/soft.rs
Freax13: Yup, good catch, it's indeed not needed.

aes/src/backends/soft.rs
Freax13: Done

.github/workflows/aes.yml
Freax13: Done

newpavlov (approved):
For posterity, could you add a comment (in the PR discussion) explaining why the `aes192`, `mix_columns`, and `utils` changes were necessary?

Freax13:
> For posterity, could you add a comment (in the PR discussion) explaining why the `aes192`, `mix_columns`, and `utils` changes were necessary?

The fixslice implementation uses the words like SIMD registers to encrypt multiple blocks in different lanes each at once. u32 can fit two lanes and u64 can fit four lanes. Because previously u32 with two lanes was the smallest denominator, some of the constants passed to `uniform_row` and `pack_rows` were already duplicated to fit that smallest denominator. This caused problems when adding the u16 implementation because a u16 can only fit a single lane.

There were two possible solutions: Either deduplicate the constants in the u16 implementation or adjust the code so that a single lane is the new common denominator and duplicate/quadruple the constants for the u32 and u64 implementations respectively. I chose the later approach.

If you look at the constants passed to `uniform_row` and `pack_rows`, you'll note that they previously all contained bit patterns with duplicated bits (e.g. `W::pack_rows(0x00, 0x03, 0x0f, 0x0c)` is equal to `W::pack_rows(0b00_00_00_00, 0b00_00_00_11, 0b00_00_11_11, 0b00_00_11_00)` and now those bit patterns have been deduplicated (the same line was changed to `W::pack_rows(0b0_0_0_0, 0b0_0_0_1, 0b0_0_1_1, 0b0_0_1_0)` or `W::pack_rows(0x0, 0x1, 0x3, 0x2)` in hex).

PR #597, opened 2026-09-17
aes: remove references to old aes_backend cfgs

### LINKED ISSUE #596: aes: `--cfg aes_backend` documentation is no longer accurate
`aes/lib.rs` still mentions setting the `aes_backend` cfg to `avx256` or `avx512`, but these options were removed in #580.

https://github.com/RustCrypto/block-ciphers/blob/24490274a3bcc91aa4cb310a332c0a17aa6d8a53/aes/src/lib.rs#L48-L49

I'd happily open a PR to fix this, but one thing I'm not sure about is what to do about the use of this flag in CI, where you have some tests that were previously explicitly forcing the VAES backends, and now auto-detect them like anywhere else:

https://github.com/RustCrypto/block-ciphers/blob/24490274a3bcc91aa4cb310a332c0a17aa6d8a53/.github/workflows/aes.yml#L105-L112

TechnoPorg:
Closes #596

cc @tarcieri


.github/workflows/aes.yml
```diff
@@ -143,7 +143,7 @@ jobs:
         include:
           - target: x86_64-unknown-linux-gnu
             rust: stable
-            RUSTFLAGS: '-Dwarnings --cfg aes_backend="avx512"'
+            RUSTFLAGS: "-Dwarnings"
```
tarcieri: Without `aes_backend` these two CI configs duplicate each other, so one can be removed.

.github/workflows/aes.yml
```diff
@@ -102,14 +102,14 @@ jobs:
...
-  vaes256:
+  vaes:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         include:
           - target: x86_64-unknown-linux-gnu
             rust: stable
-            RUSTFLAGS: '-Dwarnings --cfg aes_backend="avx256"'
+            RUSTFLAGS: "-Dwarnings"
```
newpavlov, line 112: As noted above, this line can be removed.

.github/workflows/aes.yml
newpavlov, line 132: It would be nice to have two separate SDE configurations: with and without AVX-512 to test both vaes256 and vaes512 backends.

TechnoPorg: I've made them use Alder Lake (for vaes256) and Granite Rapids (for vaes512) respectively. If you'd prefer to keep it using -future instead of pinning to a specific microarchitecture that works too

newpavlov (approved):
Thanks!

PR #598, opened 2026-09-18
aes: add inline attributes

stepantubanov:
Just a tiny perf fix. Perf was ~1.5-2x worse after updating aes from 0.9.2 to 0.9.3 and turned out it was due to some functions not getting inlined across crate boundaries.


newpavlov (approved):
Good catch! I previously fixed inlining in #575, but the attributes got lost again during refactoring in #580.

Although it's a bit surprising that such tiny functions do not get inlined (probably because of some bad interaction with `#[target_feature(enable = "...")]`).

